PR #11, opened 2017-01-11
Nagios

oybed:
#### What does this PR do?
Added support for setting up nagios (server) and nagios targets for monitoring of servers/VMs/Instances and their services/resources. **Note**: This is the initial delivery of support for setting up Nagios, and additional services/resources to be monitored, and plugins, can be added as needed. 

See the README.md in the playbook/nagios directory. 

N/A

cc: @redhat-cop/casl @etsauer


oybed:
@etsauer ping

playbooks/nagios/README.md
```diff
@@ -0,0 +1,49 @@
...
+[infra]
+dns.infra.example.com nagios_services=dns
+nfs.infra.example.com nagios_services=nfs
+
+[infra:vars]
+hostgroup_name=infra-servers
+hostgroup_alias=Infrastructure Servers
+
+# OpenShift 3 environment
+[openshift]
```
etsauer: @oybed can we set this up such that we can include the targets portion in a run of the end-to-end playbook? perhaps embed a nagios option into the post-install.yml?

oybed: @etsauer yes, I totally agree with that, but I have a lot of upcoming changes to this, so I'd like to get this initial submit done and then we can do follow-up PRs to incorporate those sort of things. OK?

Issue #1, opened 2016-10-12
Update README after repository migration


sabre1041:
Update the repository README to reference the correct repository and properly callout the contents

PR #16, opened 2017-02-07
Adding documentation and docker containers so others can begin testin…

etsauer:
…g cluster provisioning

#### What does this PR do?
Adds README content for dynamic inventory based provisioning. Also adds docker client container for running provisioning jobs

Follow README instructions.

https://github.com/redhat-cop/casl-ansible/issues/1

cc: @redhat-cop/casl

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,13 +10,14 @@ In addition to _cloning this repo_, you'll need the following:
 
 * Access to an OpenStack environment using an link:http://docs.openstack.org/user-guide/common/cli-set-environment-variables-using-openstack-rc.html[OpenStack RC File]
   * File should be placed at `~/.config/openstack/openrc.sh`
+>**NOTE**: OpenStack environment currently requires HEAT to be enabled, and the user must have the `heat_stack_owner` role assigned.
 * A link:https://github.com/naturalis/openstack-docs/wiki/Howto:-Creating-and-using-OpenStack-SSH-keypairs-on-Linux-and-OSX[Key-pair created in OpenStack]
 * Docker installed (`yum install -y docker` on RHEL/Centos, `dnf install docker -y` on Fedora)
   * If you plan to run docker as yourself (non-root), your username must be added to the `docker` user group.
 * An `~/.ansible.cfg` file containing the following:
 ```
 [defaults]
-roles_path= /root/repository/rhc-ose/rhc-ose-ansible/roles:/root/repository/openshift-ansible/roles
+roles_path=/root/repository/casl-ansible/roles:/root/repository/openshift-ansible/roles
 filter_plugins= /usr/share/ansible_plugins/filter_plugins:/root/repository/openshift-ansible/filter_plugins
 host_key_checking = False
 ```
@@ -32,7 +33,7 @@ Cool! Now you're ready to provision OpenShift clusters
 
 ### Provision a Cluster
 
-To start, we'll provision a the `sample.casl.example.com` cluster defined in the `casl-ansible/inventory` directory.
+To start, we'll provision the `sample.casl.example.com` cluster defined in the `casl-ansible/inventory` directory.
 
 1. Edit `casl-ansible/inventory/sample.casl.example.com/hosts` and edit the `# OpenStack Provisioning Variables` and `# Subscription Management` sections at the top to match your environment/project/tenant.
 2. Start the `openstack-client` container.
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -2,7 +2,47 @@
...
+
+* Access to an OpenStack environment using an link:http://docs.openstack.org/user-guide/common/cli-set-environment-variables-using-openstack-rc.html[OpenStack RC File]
+  * File should be placed at `~/.config/openstack/openrc.sh`
+* A link:https://github.com/naturalis/openstack-docs/wiki/Howto:-Creating-and-using-OpenStack-SSH-keypairs-on-Linux-and-OSX[Key-pair created in OpenStack]
+* Docker installed (`yum install -y docker` on RHEL/Centos, `dnf install docker -y` on Fedora)
+  * If you plan to run docker as yourself (non-root), your username must be added to the `docker` user group.
+* An `~/.ansible.cfg` file containing the following:
+```
+[defaults]
+roles_path= /root/repository/rhc-ose/rhc-ose-ansible/roles:/root/repository/openshift-ansible/roles
```
oybed: Looks incorrect - since the repo is now `casl-ansible`, that's what the first path should be pointed to. Also, would be good to add a note in this area to say that if the default location of the clone is not used, this needs to be updated as well.

README.md
```diff
@@ -2,7 +2,47 @@
...
+git clone git@github.com:redhat-cop/casl-ansible.git
+git clone git@github.com:openshift/openshift-ansible.git
+```
+* Copy `casl-ansible/inventory/sample.casl.example.com.d/inventory/clouds.yaml` to `~/.config/openstack/clouds.yaml`
+
+Cool! Now you're ready to provision OpenShift clusters
+
+### Provision a Cluster
+
+To start, we'll provision a the `sample.casl.example.com` cluster defined in the `casl-ansible/inventory` directory.
```
oybed: typo - "a the"?

README.md
```diff
@@ -2,7 +2,47 @@
...
+
+1. Edit `casl-ansible/inventory/sample.casl.example.com/hosts` and edit the `# OpenStack Provisioning Variables` and `# Subscription Management` sections at the top to match your environment/project/tenant.
```
oybed: Do we want to include any additional info re: additional parameters that can/should be adjusted?

README.md
```diff
@@ -2,7 +2,47 @@
...
+2. Start the `openstack-client` container.
+```
+./casl-ansible/docker/openstack-client-centos/run.sh --repository=/path/to/repos/dir/
+```
+3. Run the `end-to-end` provisioning playbook
+```
+ansible-playbook -i /root/repository/casl-ansible/inventory/sample.casl.example.com.d/inventory /root/repository/casl-ansible/playbooks/openshift/end-to-end.yml
```
oybed: Is the "openshift-ansible" path no longer needed? I believe we typically passed that as an environment variable?

README.md
```diff
@@ -35,23 +75,21 @@ The following are a list of Ansible playbooks
 
 ### Prerequisites
 
-A client environment with the appropriate entitlements to Red Hat Enterprise 7 and OpenShift Enterprise 3 is required to provision new OpenShift environments. A [host builder docker image](../docker/openshift-host-builder) is available to provide an execution environment. 
+A client environment with the appropriate entitlements to Red Hat Enterprise 7 and OpenShift Enterprise 3 is required to provision new OpenShift environments. A [host builder docker image](../docker/openshift-host-builder) is available to provide an execution environment.
 
 ### Configuring the Inventory
 
 The current supported method of provisioning new OpenShift environments is to utilize OpenStack. The `openstack_key_name` in the `ose3` inventory file at the root of this folder must have the SSH key name specified in order for Ansible to be able to communicate with the newly provisioned machines
```
oybed: Should include a note about Heat and that the OSP env need heat enabled (and potentially which heat group the user needs to be part of for the tenant - i.e.: `heat_stack_owner`)?

README.md
```diff
@@ -35,23 +75,21 @@ The following are a list of Ansible playbooks
...
 
 ### Running the OpenShift Playbook
- 
+
 Execute the following command to run the playbook
```
oybed: is this still valid? - seems a bit out of context now ...

README.md
```diff
@@ -35,23 +75,21 @@ The following are a list of Ansible playbooks
...
 
     ansible-playbook -i inventory/ose-provision ose-provision.yml
-    
+
 
 ### Running the CICD Playbook
- 
+
 Execute the following command to run the playbook
```
oybed: same as comment above

etsauer:
I have always disagreed with this idea that the Docker image that's needed to run this code shouldn't live in this repo because it's not ansible code. To me, that's like saying that my Pom file shouldn't live in a Java project because it's not ansible code. It's the thing that I need to run the stuff in this project and therefore we should keep it with this project.

If I lose this argument, then I guess we need to find another place to put it, because I don't want to create a dependency on a repository we are deprecating.

@oybed @sabre1041 I would like weigh-ins from both of you on this.

README.md
etsauer: No, openshift_ansible_path is no longer needed to be passed by the user, per PR #14

etsauer:
@oybed other than my comment above around docker images, I believe I have addressed the comments you made above. Please give another look.

README.md
oybed: Ok, missed that change. Although it's good, we'd want to ensure that users understand that `openshift-ansible` exists in that location. Yes, the clone command is shown, but not sure it necessarily calls out this "dependency" and __required__ location. Probably good enough for now unless you can think of a way to call this out.

oybed:
@etsauer re: docker files for the images - yeah, maybe best discuss next time we meet, but IIRC these docker images weren't moved originally due to wanting to find the "best home" for them. I'm fine keeping them here for now, but let's discuss.

README.md
oybed: sorry, see you got that already - looks good

PR #20, opened 2017-02-11
Defaulting the DNS instance to a small flavor

### LINKED ISSUE #19: Heat Stack deployment should NOT use default flavor for all instances
For example, right now the DNS server is using the same flavor as everything else - this often means using a lot of resources for something that requires very little.

oybed:
#### What does this PR do?
Changing the default provisioning of the DNS flavor to use m1.small to save resources. If a different flavor is desirable, the `openstack_dns_flavor` can be set in the inventory. 

Perform a deployment of the full stack (see repo README for quickstart). Validate that the DNS instance is running and performing correctly.

resolves #19 

cc: @redhat-cop/casl @etsauer


etsauer:
lgtm

PR #21, opened 2017-02-11
Updating to development release of ansible 2.3.0 to pull down bug fix…

etsauer:
…es in HEAT module

#### What does this PR do?
Pulls in ansible 2.3.0 from the development release. This contains an important fix to the `os_stack` module that allows for idempotency of the infrastructure. This, in turn gives us the ability to scale openshift clusters up and down.

Couple of notes on this:
- I've only updated the centos image for now, as this is temporary until ansible 2.3.0 is released.
- Currently, if infranodes are scaled out the router will not scale with them. I've submitted a fix upstream for this: https://github.com/openshift/openshift-ansible/pull/3340

Rebuild the centos docker image, then run a provisioning job from instructions from the README.

cc: @redhat-cop/casl

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,26 @@ To start, we'll provision the `sample.casl.example.com` cluster defined in the `
 ansible-playbook -i /root/repository/casl-ansible/inventory/sample.casl.example.com.d/inventory /root/repository/casl-ansible/playbooks/openshift/end-to-end.yml
 ```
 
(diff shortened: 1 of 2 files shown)


etsauer:
@oybed are we waiting for the official release of 2.3.0?

oybed:
Now that we have a v1.0.0 release, I'm good to merge.

PR #24, opened 2017-02-15
Ensure DNS configuration has wildcards set for infra nodes

### LINKED ISSUE #23: DNS configuration of heat based provisioning doesn't include wildcard records
It is expected that the provisioning sets up wildcard record(s) for the provisioned environment, e.g.: *.apps.example.com ==> infra node(s)

oybed:
#### What does this PR do?
Adds back in the wildcard for the DNS configuration

Run a provisioning following the instructions in the top level README. Check that the wildcard DNS records work afterwards (+ check the DNS config to validate that they are there)

resolves #23

cc: @redhat-cop/casl @etsauer


etsauer:
@oybed tested changs, and I do now get an apps subdomain. however, the DNS server does not get a record created for itself. This would be useful for ssh-ing into the hosts that get created. Can we make that change as part of this also?

oybed:
@etsauer agreed - will make that change

oybed:
@etsauer changes made to pick up all cluster hosts for DNS configuration - please review

etsauer:
@oybed looks like there are merge conflicts now.

oybed:
@etsauer merge conflicts fixed - you should be able to merge now.

PR #28, opened 2017-03-02
Dns selinux

oybed:
#### What does this PR do?
Updated to handle SELinux correctly for DNS servers

Install a dns-server and ensure that SELinux doesn't cause problems when adding new records with nsupdate

N/A

cc: @redhat-cop/casl


oybed:
Tested and working on Fedora 25 and a full provisioning on OSP10. Merging.

PR #30, opened 2017-03-04
Openshift mgmt

oybed:
What does this PR do?

added support for pruning projects based on age in hours
How should this be manually tested?

See README in the openshift-management role
Is there a relevant Issue open for this?

N/A
Who would you like to review this?

cc: @redhat-cop/casl @sabre1041


oybed:
@sabre1041 new version ... still interested in finding a better solution instead of the `item.item.item` part, so if you have any suggestions, please let me know

roles/openshift-management/tasks/prune-projects.yml
```diff
@@ -0,0 +1,39 @@
...
+  environment:
+    KUBECONFIG: "{{ kubeconfig }}"
+  register: projects 
+
+- name: Get OpenShift project timestamps
+  shell: oc {{ run_as | default('') }} get project {{ item }} --template={{ '{{' }}.metadata.creationTimestamp{{ '}}' }}
+  environment:
+    KUBECONFIG: "{{ kubeconfig }}"
+  register: project_timestamp
+  with_items: "{{ projects.stdout_lines }}"
```
sabre1041: Had to replace with this line, otherwise an error was thrown in Ansible 2.2 when pruning projects was disabled

```
with_items: "{{ projects.stdout_lines | default([]) }}"
```

oybed: @sabre1041 hmm ... it shouldn't even hit that area if pruning projects are disabled. I just tried it in my Ansible 2.2.1.0 env, and I don't see this problem. Were you using 2.2.0.0?

oybed:
@sabre1041 the bigger issue still is that Ansible 2.2.x at random times are failing due to the "template error" ... It was working for me until I just re-tried with some additional changes and it is now again failing with the "unexpected '.'" error... Ansible 2.3 seems to be working fine, however, so looking forward to using that, but for now we probably need to revert back to what I had before.

```
 :
 template error while templating string: unexpected '.'. String: oc  get project test --template={{.metadata.creationTimestamp}}"}
 :
```

sabre1041:
@oybed a few points

* Upgraded to ansible 2.2.1.0 and no longer hitting the roadblock that i had hit previously when pruning projects was not enabled
* `item.item.item` is basically what we will need to use. A little ugly, but it will do
* When running on OSX, date commands are blowing up due to differences in date syntax. If we are not targeting OSX as the the platform, we can ignore but probably should note it in the `README`

I'm good to merge after addressing the date questions

oybed:
@sabre1041 README updated with an additional Notes section. Check it out.

sabre1041:
LGTM thanks!

PR #42, opened 2017-05-05
Add an optional custom pre-install playbook

tomassedovic:
#### What does this PR do?

This adds a way to specify an optional playbook that will be run at the end of the
pre-install section (i.e. after the openstack resources are up but
before the openshift installation).

It's useful for uploading the LDAP CA to the master nodes and any other custom pre-install configuration.

I'm sot sure this is necessarily the best way to go, but it's very simple and flexible. I haven't updated the docs or anything in case this gets closed/done differently yet.

Write a new playbook that does something on the nodes (say copies a file over). Run the end-to-end deployment with `-e openstack_custom_pre_install_playbook=<path to the playbook>`. Once it completes, verify that the custom playbook was executed.

No

cc: @redhat-cop/casl


tomassedovic:
I hadn't realised that openshift-ansible already supports uploading LDAP CA with the `openshift_master_ldap_ca` and `openshift_master_ldap_ca_file` options.

I'll close this PR.

PR #43, opened 2017-05-10
Use the correct variable name in create_users

tomassedovic:
The user creation was failing, because it was looking for the
`demo_users` variable while the samples put the data under
`create_users`.

#### What does this PR do?

It makes the `create_users` section in the sample inventory actually create the users.

Add the following to `OSv3.yaml`, run the end-to-end playbook and verify you can log in with username `rdu-user00` and password `rdu-sample`:

```yaml
create_users:
  num_users: 5
  prefix: 'rdu-user'
  passwd_file: '/etc/origin/master/htpasswd'
  password: 'rdu-sample'
```

No

cc: @redhat-cop/casl


oybed:
LGTM

PR #48, opened 2017-06-07
Update CASL to use nsupdate for DNS records

### LINKED ISSUE #8: Update DNS record generation to use nsupdate
Instead of building a file for static DNS records, we should be using nsupdate (the dns-server role already supports it) to add records. I.e.: change these places within the code:

https://github.com/redhat-cop/casl-ansible/blob/master/playbooks/openshift/pre-install.yml#L17-L24
https://github.com/redhat-cop/casl-ansible/blob/master/playbooks/openshift/pre-install.yml#L32-L34
https://github.com/redhat-cop/casl-ansible/blob/master/playbooks/openshift/dns_records.yml

oybed:
#### What does this PR do?
Updates the CASL implementation to use `nsupdate` to populate DNS records in the DNS server(s). This works with both the internally provisioned DNS server as well as external DNS server(s). 

3 Test Scenarios:

1. Run the casl-ansible as-is with a valid inventory (note the updated inventory samples) and validate that the internally provisioned DNS server works as it use to 

2. Provision a "public" DNS server to be used for external access to the cluster, and leave the inventory as-is for using the private / internally provisioned DNS server within the cluster. Set the `external_nsupdate_keys['public']` configuration in the inventory to allow the tools to update an external DNS server, then run the `casl-ansible` tooling/provisioning. Observe that the cluster functions properly and that external access can be gained using the external DNS server resolution.

3. Provision 2 external DNS servers ("public" & "private"), Configure the `external_nsupdate_keys['public']` and `external_nsupdate_keys['private']` settings in the inventory, then run the `casl-ansible` tooling/provisioning. Observe that the cluster uses the "internal" DNS server for internal resolution (private IPs) while the "external" DNS server is used for public resolution. 

resolves #8 

cc: @redhat-cop/casl @tomassedovic @cooktheryan


oybed:
Note the `dns-server` role can be used to quickly configure a DNS server with the appropriate configuration, including the necessary nsupdate keys, etc.

1. Use the [dns-server](https://github.com/redhat-cop/casl-ansible/tree/master/roles/dns-server) role to configure the DNS server
2. Obtain the nsupdate secret/algorithm from the `/var/named/<view>-<dns-domain>.key` file

etsauer:
@oybed tested all thee scenarios and for the most part, the testing went well. However, while testing scenario 3, with no dns server in the cluster, I got the following warning, which halted the install and forced me to intervene to continue:
```
TASK [Warn user about bad openshift_hostname values] **************************************************************************************************************************************************************
[Warn user about bad openshift_hostname values]
The hostname "master-0.l1.casl.rht-labs.com" for "master-0.l1.casl.rht-labs.com" doesn't resolve to an ip address owned by this host. Please set openshift_hostname variable to a hostname that when resolved on the host in question resolves to an IP address matching an interface on this host. This host will fail liveness checks for pods utilizing hostPorts, press ENTER to continue or CTRL-C to abort.:
```

playbooks/openshift/end-to-end.yml
```diff
@@ -1,6 +1,7 @@
 ---
 # Provision Openstack Instances
 - include: provision-openstack.yml
+- include: post-provision-openstack.yml
```
oybed: Sure, that's fine - I'll make the change

playbooks/openshift/pre-install.yml
```diff
@@ -1,70 +1,8 @@
 ---
-# Assign hostnames
-- hosts: cluster_hosts
-  pre_tasks:
-  - include: roles/common/pre_tasks/pre_tasks.yml
-  roles:
-  - role: hostnames
-
-# Subscribe Hosts
```
oybed: The host/instance used for DNS needs to be subscribed in order to set up the necessary RPMs use by bind/named. Hence, this needs to happen before the DNS steps - i.e.: same order as before. We can, however, look into splitting them out and do a subscription just for the DNS instance here, and allow for a generic subscription approach in the pre-install steps still.

oybed:
@etsauer re: the 3rd test scenario - did you actually use 2 separate DNS servers? I checked the cluster and it doesn't look that way. Let's discuss Monday and take a closer look.

etsauer:
@oybed ah yeah. I tested with two and it looks good.

oybed:
@etsauer added `python-dns` to the Dockerfile to make sure the control host supports `nsupdate`

etsauer:
@oybed re-tested and all looks good to me.

PR #49, opened 2017-06-23
Introduce Ansible Galaxy to handle repo dependencies 

oybed:
#### What does this PR do?
This changes the setup of the CASL repo to handle the inter-repo dependencies. I.e.: in this first iteration, it pulls in dependent roles from `infra-ansible` as part of the setup prep for running CASL.

Follow the README file and ensure the `ansible-galaxy` step is executed before running the provisioning. Run the provisioning and ensure everything works as it used to. 

N/A

cc: @redhat-cop/casl


oybed:
Note that the following `ERROR` may be observed when running `ansible-galaxy`. From what I can tell, this can be safely ignored as the main goal is met with having the correct roles pulled in. Will continue to troubleshoot this later on. A quick google search indicates that this may be a bug with some versions of Celery and/or Ansible.

```
ansible-galaxy install -r casl-requirements.yml -p roles
- extracting infra-ansible to roles/infra-ansible
- infra-ansible was installed successfully
ERROR! Unexpected Exception: 'NoneType' object has no attribute 'get'
to see the full traceback, use -vvv
```

Will troubleshoot this later on.

sabre1041:
@oybed Opened a new [PR](https://github.com/redhat-cop/infra-ansible/pull/34) against `infra-ansible` to address the _NoneType_ error that you are seeing when importing roles

oybed:
@sabre1041 thx - it has been merged. I'd still argue there are some bugs there, but this workaround should be sufficient for what we need. Thx again.

README.md
```diff
@@ -39,8 +39,16 @@ ln -fs openshift-ansible-*<version>* openshift-ansible
 
 | openshift-ansible url     | OpenShift version | 
 |:-------------------------:|:-----------------:|
+| [3.5.77-1](https://github.com/openshift/openshift-ansible/archive/openshift-ansible-3.5.77-1.tar.gz) | OCP 3.5.x |
 | [3.4.50-1](https://github.com/openshift/openshift-ansible/archive/openshift-ansible-3.4.60-1.tar.gz) | OCP 3.4 |
 
+* Run `ansible-galaxy` to pull in the necssary requirements for the CASL provisioning:
```
etsauer: @oybed prior to this change, if the user was using our container image to run this installs, there was no requirement to actually have ansible installed on localhost. With ansible-galaxy, it looks like we are adding a requirement for ansible. Is that true, and are we ok with that? If so, we need to document in the README, that the user needs to install `ansible` and which versions we support.

README.md
```diff
@@ -39,8 +39,16 @@ ln -fs openshift-ansible-*<version>* openshift-ansible
...
+
+```
+cd ~/src/casl-ansible
+ansible-galaxy install -r casl-requirements.yml -p roles
+```
+
 Cool! Now you're ready to provision OpenShift clusters
 
 ### Provision a Cluster
```
etsauer: While we're in here, can we add to this section something about the options for setting up DNS? right now, with the galaxy integration we've done, if I run a normal provisioning job that builds a cluster with DNS, then i won't actually exercise the role that was moved over.

README.md
oybed: @etsauer it can all be done from the control host - i.e.: no need to run it from localhost. I think we need to rework a major portion of this README file anyway as it's seems to be confusing to users, so maybe we can look at figure out how to better structure this as a follow-up PR (including how to balance information in the top level README v.s. the control host README).

README.md
oybed: @etsauer not sure I quite follow - it seems that you may have it backwards. I.e.: if you choose to run a "normal" provisioning job that builds a cluster with DNS, it will **indeed** exercise the role(s) that were moved over. In either case, it will use one or both of the roles currently moved. Anyway; maybe this can be part of the bigger revamp that needs to be happen for this README to restructure / clarify the usage?

casl-requirements.yml
```diff
@@ -0,0 +1,8 @@
+
+# This is the Ansible Galaxy requirements file to pull in the correct roles
+# to support the operation of CASL provisioning/runs. 
+
+# From 'infra-ansible'
+- src: https://github.com/redhat-cop/infra-ansible
```
bogdando: How could I test this?
I have an error:
```
- downloading role from src: https://github.com/redhat-cop/infra-ansible
 [ERROR]: failed to download the file: <urlopen error unknown url type: src>
```

bogdando: Although the command ``ansible-galaxy install git+https://github.com/oybed/infra-ansible,galaxy -p roles`` works

oybed: Hmm ... maybe an environment issue on your localhost? Everything seems to be working fine as-is here:
```
$ cat requirements.yml 
- src: https://github.com/redhat-cop/infra-ansible

$ ansible-galaxy install -r requirements.yml -p roles
- extracting infra-ansible to roles/infra-ansible
- infra-ansible was installed successfully

$ ls -l roles/
total 4
drwxrwxr-x. 6 obedin obedin 4096 Jul  3 10:05 infra-ansible

$ ansible-galaxy --version
ansible-galaxy 2.3.1.0
  config file = /etc/ansible/ansible.cfg
  configured module search path = Default w/o overrides
  python version = 2.7.13 (default, May 10 2017, 20:04:28) [GCC 6.3.1 20161221 (Red Hat 6.3.1-1)]
```

bogdando: I'm using
ansible-galaxy 2.3.1.0, python version = 2.7.12

PR #60, opened 2017-08-15
Adding -n to oc apply

logandonley:
#### What does this PR do?
It adds the namespace flag to the oc apply command which means that more templates can be used without having to edit in the namespace parameter for every object. 

This adds a new requirement that all templates will need to include a params file which at minimum includes the NAMESPACE parameter. 

There is a test playbook in the test directory for the oc-apply role.

N/A

cc: @redhat-cop/casl @oybed


oybed:
@logandonley this looks good. However, it doesn't eliminate the need for the NAMESPACE parameter in the template (part of the `parameters` at the end of the template file). Basically using `oc process` with the param file containing the `NAMESPACE` parameter means it will require the template to have it in the parameters list as well. It does, however, eliminate the need to have it added to every object within the template.

oybed:
@logandonley also, one more observation - if the `NAMESPACE` parameter is missing from the parms file, it will will error out, but the error message isn't as clear as it can be - i.e.: `oc apply` errors with "namespaces \"[]\" not found. We should submit a follow-up PR to clarify this error.

Issue #70, opened 2017-08-22
openshift-applier: clean-up errors for URL checks


oybed:
As part of the task to check if a passed in template is a url or local file, we should use `failed_when` to avoid printing red failure messages, or even consider something like `method: HEAD` with the `url` module (as described [here](https://stackoverflow.com/questions/42485260/ansible-check-file-exists-in-the-http-path-before-get-url-win-get-url).

oybed:
Done - closing

PR #71, opened 2017-08-23
Adding failed_when check to make sure it doesn't fail for local files

logandonley:
#### What does this PR do?
Previously, the "Check if the passed in template is a remote location (URL)" task need `ignore-errors: yes` since it would fail if the template wasn't a remote location.

This PR adds a failed_when check which will make it so it doesn't fail when it is a local template or a oob template.

Run the test playbook included with the role. Ideally, the inventory you include would use a remote template, a local template, and an oob template. 

[#70](https://github.com/redhat-cop/casl-ansible/issues/70)

cc: @redhat-cop/casl @oybed


oybed:
@logandonley good job - merging!!

Issue #64, opened 2017-08-16
openshift-applier: Document the usage with a proper README file


oybed:
The `oc-apply` role right now is very much undocumented - i.e.: we need a detailed description of the inventory and its dictionary definition.

oybed:
Done - closing

PR #72, opened 2017-08-23
Adding role variables documentation

logandonley:
#### What does this PR do?
Adding documentation on the variables for the oc-apply role.

N/A

[#64](https://github.com/redhat-cop/casl-ansible/issues/64)

cc: @redhat-cop/casl @oybed

### DIFF
--- a/roles/oc-apply/README.md
+++ b/roles/oc-apply/README.md
@@ -12,7 +12,87 @@ A working OpenShift cluster that can be used to populate things like namespaces,
 Role Variables
 --------------
 
-TBD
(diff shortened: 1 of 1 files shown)


logandonley:
Let me know if there are any areas you want me to change/expand on.

oybed:
@logandonley First quick scan looks good - I'll provide more detailed feedback a bit later. However, one thing I believe is missing is the `content_dir` - can you please look to add something about it?

https://github.com/redhat-cop/casl-ansible/blob/master/roles/oc-apply/tasks/apply-content.yml#L14-L19

Here's an example inventory that has it:
https://github.com/oybed/casl-ansible/blob/demo/inventory/sample.casl.example.com.d/inventory/group_vars/all.yml#L73

logandonley:
@oybed Good call, I'll add a section for that.

PR #73, opened 2017-09-05
Role to add OpenShift container provider to CloudForms

sabre1041:
#### What does this PR do?

1. Have a CloudForms instance available
2. Use example playbook from PR substituting values as necessary

No

cc: @redhat-cop/casl @oybed


makentenza:
@sabre1041 just checked this. It works fine. A couple of comments:

-  Why don't add automatically the token from management-admin SA in management-infra namespace?
- Use FQN for Kubernetes services like kubernetes.default.svc.cluster.local and hawkular-metrics.openshift-infra.svc.cluster.local. I think is a good practice and I have seen errors in the past as Node resolver was not properly configured with domain 'search' options

sabre1041:
@makentenza 

* Updated hostnames to include full DNS address 
* On the fence on how to utilize the token from the SA preconfigured in the cluster. Maybe only use it if another one is not provided? There may be situations where a user may create an SA for this purpose and want to use it instead. Thoughts?

makentenza:
@sabre1041, yes, it makes sense to me as well to use the preconfigured SA from the cluster only if no one is provided. All the rest LGTM, I've tested and it worked fine.

sabre1041:
@makentenza sounds good. I'll add in that type of functionality and follow up for another review afterward

sabre1041:
@makentenza I added the functionality to utilize the default service account token. Please review at your convenience

makentenza:
@sabre1041, reviewed. The functionality works fine, but please update the README.md to reflect 'ocp_token' is no longer mandatory and add some explanation as well so anyone that consume the role understands what SA is using and what are the alternatives. Apart of that is ready to merge from my point of view.

sabre1041:
@makentenza added README content

makentenza:
Cool @sabre1041, LGTM, but I have no permissions to merge this PR.

oybed:
LGTM - thanks @makentenza for working with @sabre1041 on getting this reviewed / tested. Merging.

PR #75, opened 2017-09-05
Adding applier to standard post-install phase

etsauer:
#### What does this PR do?
See above

Run post-install.yml with and without `openshift_cluster_content` set.

n/a

cc: @redhat-cop/casl


oybed:
LGTM - merging

PR #78, opened 2017-09-05
Moving 'openshift-login' from 'ansible-stacks'

oybed:
#### What does this PR do?
Moving the 'openshift-login' role (with the corresponding 'openshift-defaults' role) from 'ansible-stacks' to make it available for use within the `casl-ansible` repo (and anything leveraging 'casl-ansible')

N/A - just a roles move

N/A

cc: @redhat-cop/casl


oybed:
@etsauer we'd like to keep the history here, so please make sure to use "merge commit" and not squash the history.

PR #79, opened 2017-09-06
Removing nagios roles (after move to infra-ansible)

oybed:
#### What does this PR do?
Removing nagios roles as they have been moved to `infra-ansible` - see [this PR](https://github.com/redhat-cop/infra-ansible/pull/71)

Mostly just moving code, but the playbook can be tested with the dependency on `infra-ansible` once the above mentioned PR has been merged. 

N/A

cc: @redhat-cop/casl


oybed:
This is dependent on having PR https://github.com/redhat-cop/infra-ansible/pull/71 merged.

PR #80, opened 2017-09-06
Role to deploy containerized CloudForms to OpenShift

sabre1041:
### What does this PR do?
Deployment of the containerized version of CloudForms to OpenShift

### How should this be tested?
Follow the example playbook in the README

### Is there a relevant Issue open for this?
No

cc: @redhat-cop/casl  @oybed


makentenza:
@sabre1041, just tested and even it works fine I have a few comments:

- I understand why you don't merge both deploy CF and config CF and it's great to have a role to add OCP provider to an external CF, but why don't include that part as well in this role to fully automate the process so the role would not just deploy CF but also configure it? Just an idea
- I think we should include the required PVs for this in the README requirements part. The template will auto create the PVCs but you still need available PVs to be used
- I should include the CF admin credentials in the README so anyone that is not aware of the default password for CF will not need to find it
- Could be possible to remove the "dirty" ouput of querying for the 'cloudforms' namespace using 'failed_when' and 'no_log'?

oybed:
@sabre1041 Can we consider using the `openshift-applier` approach with this as well? From what I can tell, there's nothing (besides of the "wait for CF to become available) that's not support by the `openshift-applier` as-is. Thoughts?

@makentenza in general, we'd like to keep the roles generic and not do "too much" for easy re-use, then use playbooks to stitch them together to be the complete implementations you're referring to. For these particular roles, I like the split @sabre1041 has here as it allows us to re-use the config role for the appliance (i.e.: non-OpenShift hosted) deployment of CF as well.

oybed:
@makentenza yup, that's what the playbook can do - potentially a follow-up PR to add the playbook unless @sabre1041 can include it with one of these PRs.

makentenza:
@sabre1041 there are some roles we should activate on the deployed CF:

Capacity & Utilization Coordinator
Capacity & Utilization Data Collector
Capacity & Utilization Data Processor

oybed:
@sabre1041 @makentenza re: Mak's latest comment re: CF Roles that should be activated - wouldn't that be a better fit for the `cfme-ocp-provider` role? I.e.: as part of setting up OpenShift as a provider, also enable the listed CF roles that Mak mentioned?

makentenza:
@oybed not really. Those roles must be activated on any or many workers, on different Zones, and these are not just for OCP. So the role that adds the OCP provider doesn't really know how your CF deployment is and it should be part of the CF administrator to take care about that.

It make sense to activate those roles here as we are using this CF initially just for OCP, so we need those roles activated in order to CF collect data and create charts.

oybed:
@makentenza So this particular PR (and the role introduced) is to deploy (i.e.: install CF) on top of OpenShift. The CF roles that should be activated, should be activated regardless of where CF is running. Hence, my suggestion to do those in a separate roles so we don't have to repeat for other roles that are created to deploy CF on top of other infrastructure (e.g: deploying the appliance on top of a VM infrastructure). Basically, these are steps that needs to be performed regardless of how and where CF is running, hence I don't feel it is correct to add these steps to this particular PR/role.

oybed:
@sabre1041 with v3.7, is this PR still needed?

sabre1041:
@oybed no. 3.7 provides the functionality to deploy CFME within OCP. Closing

PR #81, opened 2017-09-06
Updated to handle remote execution with local inventory files

oybed:
#### What does this PR do?
Added functionality to copy the local files (part of the inventory) to a remote host where the `oc` commands are executed. 

#### How should this be manually tested?
Update the inventory in the tests directory to use a remote host (e.g.: an OpenShift master host) and run the test. 

#### Is there a relevant Issue open for this?
N/A

#### Who would you like to review this?
cc: @redhat-cop/casl

PR #83, opened 2017-09-19
Adjusting seed cluster step to take advantage of system:admin on the …

etsauer:
…master

#### What does this PR do?
runs the openshift-applier role from first master.

Run through `playbooks/openshift/end-to-end.yml` or `playbooks/openshift/post-install.yml` on an existing cluster

depends on https://github.com/redhat-cop/casl-ansible/pull/81

cc: @redhat-cop/casl

### DIFF
--- a/playbooks/openshift/post-install.yml
+++ b/playbooks/openshift/post-install.yml
@@ -8,6 +8,8 @@
   roles:
   - { role: create_users }
 
-- include: ../openshift-cluster-seed.yml
+- hosts: masters[0]
+  roles:
+    - openshift-applier
   when:
   - openshift_cluster_content is defined


oybed:
LGTM - merging

oybed:
@etsauer actually, this seems to break functionality. What version of `ansible` did you test with that had this working? When I use version `2.3.1.0`, I get:
```
ERROR! 'when' is not a valid attribute for a Play

The error appears to have been in '/root/repository/casl-ansible/playbooks/openshift/post-install.yml': line 11, column 3, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

- hosts: masters[0]
  ^ here
```

PR #85, opened 2017-09-23
Openstack delete

oybed:
#### What does this PR do?
- Introduces a couple of playbooks to allow for easy de-commision of a OpenShift cluster (e.g.: OpenStack Stack)
- Updating inventory to avoid running everything with privileged rights - i.e.: only apply `ansible_become` to the `OSEv3` and `dns` groups

1. Provision cluster as normal (following README) to validate that it still works
2. Use the `delete-cluster` playbook to decomission an OpenShift cluster

N/A

cc: @redhat-cop/casl @etsauer @makentenza @day4skiing


playbooks/openshift/delete-cluster.yml
```diff
@@ -0,0 +1,5 @@
+---
+
+- include: delete-openstack.yml
```
day4skiing: Consider using `include_vars: "delete_{{ hosting_infrastructure }}.yml" ` instead, allowing it to fail when someone uses an unsupported infrastructure.

oybed: Hmm ... so `include_vars` isn't valid for this, so I presume you mean just a `include`? If so, that introduces some other challenges - i.e.: the variable isn't defined at this level, and if you try to wrap it in a set of `tasks`, there are other hurdles to work through. Although the idea is great, I'm not sure it as simple as it may sound. 

Also see this comment for some more details:
https://github.com/ansible/ansible/issues/15309#issuecomment-206786728

This may change with Ansible 2.4 where there's new functionality around this - i.e.: `import_play` (or `import_playbook`) - I just haven't had a chance to do much with this yet + we're currently on `2.3.x`

day4skiing: The hope was to have `delete-cluster.yml` be a "one-liner" and as more infrastructures become supported it would simply require the creation of an additional infrastructure specific file called `delete_{{ hosting_infrastructure }}.yml` without having to touch `delete-cluster.yml`.   As it is written now every additional supported infrastructure will require a change to `delete-cluster.yml` as well. 

If the `hosting_infrastructure` variable is not defined, it doesn't make sense to call `delete-cluster.yml` to me.   If that is a possible scenario then we have a "silent failure" on our hands and I would recommend the following one-liner `include_vars: "delete-{{ hosting_infrastructure | default(error) }}.yml` where `delete-error.yml` displays/logs a message indicating that the cluster was `not deleted`.

oybed: Totally understand where you are coming from, and as said above, the idea is great. I've went down that path before landing on the current solution and wish it would work. We have done similar things in the past, but this isn't a place where the proposed solution works.

Anyway; I'm thinking there may be some misunderstandings, so this is most likely better discussed offline. Let's chat Monday.

day4skiing: Thanks for the discussion this morning.  LGTM

etsauer:
@oybed before reviewing this, see my comments in the upstream PR

oybed:
@etsauer now that the `openshift-ansible-contrib` PR has been merged, can we please finish this one?

PR #93, opened 2017-09-30
adding whitelist feature

### LINKED ISSUE #90: openshift-applier: Allow for some method of only running portions of the inventory
For example, we may not always want to re-run certain segments of the inventory on every run, so a way to perhaps set a flag to only run certain objects within the inventory would be ideal.

@sherl0cks and @infosec812 have both been experiencing issues around this.

sherl0cks:
#### What does this PR do?
Enables end users to specify a variable `wl`, which a comma delimited list of objects in the inventory to process. Any object not named in the list, will not be processed. Thus, a whitelist.

README provides an example when turning the feature on. If the feature is not activated, the role will automatically add all objects to the whitelist.

resolves #90 

cc: @redhat-cop/casl

### DIFF
--- a/roles/openshift-applier/README.md
+++ b/roles/openshift-applier/README.md
@@ -93,6 +93,17 @@ These objects look like this:
 **_NOTE:_** If the target namespace is not defined in each of the objects within the template, be sure to add the `namespace` variable. 
 
 
+Executing a Subset of the Inventory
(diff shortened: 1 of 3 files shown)


oybed:
Closing this out as PR #104 takes its place instead.

PR #94, opened 2017-10-01
change process task title to be more relevant

sherl0cks:
#### What does this PR do?
Update a single task to be more description. Today, this outputs the object name, not the specific template name you are processing, which is confusing when processing an object with many templates

This isn't a functional change. Nothing to other than regression

No

cc: @redhat-cop/casl


roles/openshift-applier/tasks/process-one-entry.yml
```diff
@@ -54,7 +54,7 @@
   - item.namespace is defined
   - item.namespace|trim != ''
 
-- name: "Create OpenShift objects for '{{ entry.object }}' with params"
+- name: "Create OpenShift objects for '{{ item.name }}' with params"
```
oybed: The `name` isn't guaranteed to be set, so this could potentially generate errors.

PR #99, opened 2017-10-04
Allow for idempotent creation of specified objects

### LINKED ISSUE #98: openshift-applier: Allow some content to be "created" instead of "applied" for idempotency
When using oc-applier to create persistent databases with generate credentials, subsequent runs of the inventory will overwrite the secrets stored in OpenShift but have no effect on the database itself. I propose added an optional field in the inventory so that a particular template is applied once and only once on a given cluster. For example:

```
- object: someobjects
  content:
  - name: ephemeraldb
    template: openshift//mysql-ephemeral
    namespace: my-project
    params: "{{ inventory_dir }}/../params/ephemeraldb/build_params"
  - name: persistentdb
    template: openshift//mysql-persistent
    namespace: my-project
    params: "{{ inventory_dir }}/../params/persistentdb/build_params"
    idempotent: true
```

Inside of of this project, we create a copy of `roles/oc-apply/tasks/process-content.yml` and name it `roles/oc-apply/tasks/process-idempotent.yml`. This task will be modified to ONLY run when `idempotent` is defined and the original will be modified to ONLY run when idempotent is undefined. The new task will use `oc create` along with `oc process` instead of using `oc apply`. The new task would also swallow errors related to the objects already existing.

InfoSec812:
#### What does this PR do?
Add new task `process-idempotent` to role `oc-apply`
which can handle idempotently creating objects
in OpenShift.

Add the field `idempotent` to an object to be created from a template and run the inventory.
Run the same inventory again and verify that the object is unchanged

Resolves #98 

cc: @oybed @sherl0cks

### DIFF
--- a/roles/oc-apply/tasks/main.yml
+++ b/roles/oc-apply/tasks/main.yml
@@ -30,7 +30,7 @@
   when:
   - entry.object not in [ 'namespace', 'policy' ]
 
-- name: "Create OpenShift objects with params"
+- name: "Create OpenShift objects with params (non-idempotent)"
   include: process-content.yml
   with_items:
   - "{{ openshift_cluster_content }}"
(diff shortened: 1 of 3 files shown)


oybed:
Closing out this PR as #102 should have taken care of some of this + with the upcoming filter/tag work, it should be even more aligned with your request.

PR #100, opened 2017-10-04
Preparing CASL for a release with update contrib

oybed:
#### What does this PR do?
Updating the CASL repo to utilize the newly cut v.3.6.0 release of upstream contrib

Ensure CASL provisioning works as expected

N/A

cc: @redhat-cop/casl


README.md
```diff
@@ -1,37 +1,44 @@
...
+
 ```
 cd ~/src/
-git clone git@github.com:redhat-cop/casl-ansible.git
+git clone https://github.com/redhat-cop/casl-ansible.git
 ```
-* Copy `casl-ansible/inventory/sample.casl.example.com.d/inventory/clouds.yaml` to `~/.config/openstack/clouds.yaml`
 
-* Download/untar `openshift-ansible` (or use git clone) for use as part of the install. Make sure to do so within the same directory as above (i.e.: `~/src`), and either rename the directory to `openshift-ansible` or create symlink to it (see example below). See table below for versions / urls to be used for the download. Note that other versions may work as well, but these are the ones we have tested and found to be stable. Feel free to submit PRs for this file with updated versions as they are found to be functional.
+* Download/untar `openshift-ansible` for use as part of the install. Make sure to do so within the same directory as above (i.e.: `~/src`), and either rename the directory to `openshift-ansible` (or create symlink to it - see example below). See table below for versions / urls to be used for the download.
```
etsauer: Why did we remove the git clone portion? In my mind we should be pushing people more towards `git clone/git checkout release-3.x` than downloading tars.

README.md
```diff
@@ -1,37 +1,44 @@
 #  CASL Ansible
 
-Automation of OpenShift 3 using [Ansible](http://www.ansible.com/)
+Automation of OpenShift 3 using [Ansible](http://www.ansible.com/).
```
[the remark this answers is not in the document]
oybed: sounds good - will do

README.md
oybed: Maybe more of a preference;  I believe going the route of downloading a tar-file is "simpler" + allows for an easier way of using a particular release rather than always the "latest" 3.x.y.

PR #104, opened 2017-10-07
Introducing tags and filter_tags for partial runs of inventory 

### LINKED ISSUE #90: openshift-applier: Allow for some method of only running portions of the inventory

oybed:
#### What does this PR do?
Introducing filter & tags to allow for filtering which portion of an inventory should be applied. This PR is inspired by PR #93 and conversation in issue #90 

Add the `tags` list to some portion of the inventory, then run with the `filter_tags=<tag>,<tag>`

Resolves #90 

cc: @redhat-cop/casl @sherl0cks @InfoSec812


sherl0cks:
@oybed @etsauer I reviewed the code and I like the approach. This provides the functionality I need, and it keeps the model "ansible native" as has been the theme with the applier. I haven't reviewed the code, but so long as meets your implementation requirements, it meets my functional requirements. 

Nice job @oybed

PR #106, opened 2017-10-11
New role to support provisioning in AWS

makentenza:
#### What does this PR do?
A new role to extend the existing capabilities of supported IaaS providers adding AWS

Use the existing inventory samples (HA not tested yet, work in progress in a different branch) to create your own and then follow the existing instructions for CASL to provision new environments

Not related with any existing Issue

cc: @redhat-cop/casl


README.md
```diff
@@ -90,6 +95,10 @@ docker exec -it <container-name> bash
 ansible-playbook -i /root/code/casl-ansible/inventory/sample.casl.example.com.d/inventory /root/code/casl-ansible/playbooks/openshift/end-to-end.yml -e openstack_ssh_public_key=<your_ssh_key_name>
 ```
 
+```
+ansible-playbook -i /root/code/casl-ansible/inventory/ample.aws.example.com.d/inventory /root/code/casl-ansible/playbooks/openshift/end-to-end.yml --private-key==<your_ssh_private_key_file>
```
oybed: Not sure what we want two commands listed here as that will be confusing. Maybe you can make the existing one a bit more generic - i.e.: something like `-i <inventory>` with a comment below that shows something like : 
```
where `<inventory>` is the path to your inventory directory as mentioned above.

File: README.md [hunk 2]
```diff
@@ -112,6 +121,16 @@ and then re-running the playbook.
 ```
 ansible-playbook -i /root/code/casl-ansible/inventory/sample.casl.example.com.d/inventory /root/code/casl-ansible/playbooks/openshift/end-to-end.yml
 ```
+### Start/Stop  a Cluster
+
+In the case we are using a Cloud Provider (AWS) to run the Cluster, it may be interesting to stop/start the instances where the Cluster is running in order to reduce costs. To do so, two playbooks are available, `start-cluster.yml` and `stop-cluster.yml`, re-using the same inventory:
+
+```
+ansible-playbook -i /root/code/casl-ansible/inventory/ample.aws.example.com.d/inventory /root/code/casl-ansible/playbooks/openshift/start-cluster.yml --private-key==<your_ssh_private_key_file>
```
oybed: I think it's confusing to include the `--private-key` variable for these `AWS` related commands, but not the OpenStack ones. I'd suggest leaving it out, but maybe add a note somewhere else on the page about using the `--private-key` option if the default `id_rsa` isn't the key used

inventory/sample-ha.aws.example.com.d/inventory/ec2.ini
```diff
@@ -0,0 +1,190 @@
+#{{ ansible_managed | Ansible EC2 external inventory script settings }}
```
oybed: Why do we have both this .ini and a template for it? Could we eliminate this one and just keep the template?

inventory/sample-ha.aws.example.com.d/inventory/group_vars/all.yml
```diff
@@ -0,0 +1,56 @@
+# 'hosting_infrastructure' is used to drive the correct behavior based
+# on the hosting infrastructure, cloud provider, etc. Valid values are:
+# - 'openstack'
+# - 'aws' (Coming Soon)
```
oybed: Please remove the 'Coming Soon' + do the same in the OpenStack related inventory

inventory/sample-ha.aws.example.com.d/inventory/group_vars/all.yml
```diff
@@ -0,0 +1,56 @@
...
+# - 'azure' (Coming Soon)
+hosting_infrastructure: aws
+
+# AWS Provisioning Variables
+# use the "-e" option to set "aws_ssh_public_key"
+aws_image_name: <ec2 AMI>
```
oybed: Can we fill in an example image name here - i.e.: the RHEL image in AWS?

inventory/sample-ha.aws.example.com.d/inventory/group_vars/all.yml
```diff
@@ -0,0 +1,56 @@
...
+aws_num_masters: 3
+aws_num_nodes: 5
+aws_num_infra: 3
+aws_access_key: <Access key ID>
```
oybed: Could we just load these up from environment variables rather than having them set in the inventory? I.e.: something like: `"{{ lookup('env','AWS_ACCESS_KEY') }}"`

inventory/sample-ha.aws.example.com.d/inventory/group_vars/all.yml
```diff
@@ -0,0 +1,56 @@
...
+aws_secret_key: <Secret access key>
+aws_region: <AWS Region>
```
oybed: Can we fill in some sample values here - i.e.: don't have to be valid, but something that looks like the actual values

inventory/sample-ha.aws.example.com.d/inventory/group_vars/all.yml
```diff
@@ -0,0 +1,56 @@
...
+aws_key_name: <Key pair name>
+aws_subnet: <Subnet ID>
+# If aws_create_vpc is 'true', aws_subnet variable will be ignored
+aws_create_vpc: <true|false>
+# These are the security groups created when aws_create_vpc is 'true'. Modify accordingly to your environment in case using existing VPC and SGs
```
oybed: Could we make the role create a set of default security groups and hence don't have to specify these in every inventory?

inventory/sample-ha.aws.example.com.d/inventory/group_vars/all.yml
```diff
@@ -0,0 +1,56 @@
...
+# Cluster variables
+env_id: "sample"
+public_dns_domain: "sre.tests.com"
+public_dns_nameservers:
+- 8.8.8.8
+ha_mode: true
+
+
+# Dockr configuration variables
+#docker_volume_size: "5"
```
oybed: Why commented out? We need to source the size from somewhere?

inventory/sample-ha.aws.example.com.d/inventory/hosts
```diff
@@ -0,0 +1,40 @@
...
+
+# Set variables common for all OSEv3 hosts
+#[OSEv3:vars]
+
+# For OSEv3 normal group vars, see ./group_vars/OSEv3.yml
+
+# Host Groups
+
+[masters:children]
+tag_group_masters_aws
```
oybed: Is there a way to avoid the `tag_group` portion of these? I.e.: maybe a setting in the .ini file?

inventory/sample.aws.example.com.d/inventory/ec2.ini
```diff
@@ -0,0 +1,190 @@
+#{{ ansible_managed | Ansible EC2 external inventory script settings }}
```
oybed: Same comments as above

inventory/sample.aws.example.com.d/inventory/group_vars/all.yml
```diff
@@ -0,0 +1,56 @@
+# 'hosting_infrastructure' is used to drive the correct behavior based
```
oybed: Pretty much the same comments for this file as the HA version

roles/deploy-aws-infra/tasks/main.yml
```diff
@@ -0,0 +1,343 @@
+---
+# tasks file for deploy-infra
+- name: Create VPC if required
+  ec2_vpc_net:
+    aws_access_key: "{{ aws_access_key }}"
+    aws_secret_key: "{{ aws_secret_key }}"
+    name: "{{ env_id }}-vpc"
+    cidr_block: 172.31.0.0/16
```
oybed: Please make this a variable and default it to this value.

roles/deploy-aws-infra/tasks/main.yml
```diff
@@ -0,0 +1,343 @@
...
+    aws_access_key: "{{ aws_access_key }}"
+    aws_secret_key: "{{ aws_secret_key }}"
+    name: "labs-ocp-master"
+    description: "Labs OCP Master"
+    vpc_id: "{{ new_vpc.vpc.id }}"
+    region: "{{ aws_region }}"
+    state: "present"
+    tags:
+      env_id: "{{ env_id }}"
+    rules:
```
oybed: Can we please source this from a variable and set them in the defaults so we can easily override if needed?

roles/deploy-aws-infra/tasks/main.yml
```diff
@@ -0,0 +1,343 @@
...
+        to_port: 10250
+        cidr_ip: 0.0.0.0/0
+  when: aws_create_vpc
+  register: ocp_node_sg
+
+- name: Create OCP Infra Security Group for the new VPC
+  ec2_group:
+    aws_access_key: "{{ aws_access_key }}"
+    aws_secret_key: "{{ aws_secret_key }}"
+    name: "labs-ocp-infra"
```
oybed: Please no hardcoded value that's related to Labs or CASL

roles/deploy-aws-infra/tasks/main.yml
```diff
@@ -0,0 +1,343 @@
...
+        to_port: 22
+        cidr_ip: 0.0.0.0/0
+  when: aws_create_vpc
+  register: ssh_sg
+
+- name: Create OCP Master Security Group for the new VPC
+  ec2_group:
+    aws_access_key: "{{ aws_access_key }}"
+    aws_secret_key: "{{ aws_secret_key }}"
+    name: "labs-ocp-master"
```
oybed: Please no hardcoded values related to Labs or CASL

roles/deploy-aws-infra/tasks/main.yml
```diff
@@ -0,0 +1,343 @@
...
+        to_port: 4789
+        cidr_ip: 0.0.0.0/0
+  when: aws_create_vpc
+  register: ocp_master_sg
+
+- name: Create OCP Node Security Group for the new VPC
+  ec2_group:
+    aws_access_key: "{{ aws_access_key }}"
+    aws_secret_key: "{{ aws_secret_key }}"
+    name: "labs-ocp-node"
```
oybed: no hardcoded values related to Labs (or CASL)

roles/deploy-aws-infra/tasks/main.yml
```diff
@@ -0,0 +1,343 @@
...
+
+# Create local facts with new subnet id
+- name: Register new created objects as facts
+  set_fact:
+    aws_subnet: "{{ new_subnets.subnet.id }}"
+
+- name: Create Master and etcd instances
+  ec2:
+    image: "{{ aws_image_name }}"
+    instance_type: m4.xlarge
```
oybed: Let's use a variable for the flavor.

roles/deploy-aws-infra/tasks/main.yml
```diff
@@ -0,0 +1,343 @@
...
+    aws_access_key: "{{ aws_access_key }}"
+    aws_secret_key: "{{ aws_secret_key }}"
+    region: "{{ aws_region }}"
+    vpc_subnet_id: "{{ aws_subnet }}"
+    exact_count: 1
+    count_tag:
+      Name: "{{ env_id }}-ocp-master-{{ item }}"
+    instance_tags:
+      Name: "{{ env_id }}-ocp-master-{{ item }}"
+      user: sre
```
oybed: Let's use a variable instead of hardcoded `sre`

roles/deploy-aws-infra/tasks/main.yml
```diff
@@ -0,0 +1,343 @@
...
+      role: ocp
+      group: masters_aws
```
oybed: Let's include the `env_id` here to avoid name clashes

roles/deploy-aws-infra/tasks/main.yml
```diff
@@ -0,0 +1,343 @@
...
+      node_labels: '{"region": "default"}'
+      ocp_role: "{{ master_tag }} {{ etcd_tag}}"
```
oybed: Can we enhance this to support both combined master/etcd as well as separate etcd ?

roles/deploy-aws-infra/tasks/main.yml
```diff
@@ -0,0 +1,343 @@
...
+      env_id: "{{ env_id }}"
+  with_sequence:
+    count="{{ aws_num_masters }}"
+  register: instances_master_etcd
+
+- name: Create Infra instances
+  ec2:
+    image: "{{ aws_image_name }}"
+    instance_type: i3.xlarge
```
oybed: variable please

roles/deploy-aws-infra/tasks/main.yml
```diff
@@ -0,0 +1,343 @@
...
+    aws_access_key: "{{ aws_access_key }}"
+    aws_secret_key: "{{ aws_secret_key }}"
+    region: "{{ aws_region }}"
+    vpc_subnet_id: "{{ aws_subnet }}"
+    exact_count: 1
+    count_tag:
+      Name: "{{ env_id }}-ocp-infra-{{ item }}"
+    instance_tags:
+      Name: "{{ env_id }}-ocp-infra-{{ item }}"
+      user: sre
```
oybed: Please use variable instead of `sre`

roles/deploy-aws-infra/tasks/main.yml
```diff
@@ -0,0 +1,343 @@
...
+      role: ocp
+      group: infra_aws
+      node_labels: '{"region": "infra"}'
```
oybed: Let's use a variable for the node_labels

roles/deploy-aws-infra/tasks/main.yml
```diff
@@ -0,0 +1,343 @@
...
+  with_sequence:
+    count="{{ aws_num_masters }}"
+  register: instances_master_etcd
+
+- name: Create Infra instances
+  ec2:
+    image: "{{ aws_image_name }}"
+    instance_type: i3.xlarge
+    group: "{{ aws_infra_sgroups }}"
+    termination_protection: yes
```
oybed: Let's use a variable for `termination_protection`, but default it to `yes`

roles/deploy-aws-infra/tasks/main.yml
```diff
@@ -0,0 +1,343 @@
...
+    aws_access_key: "{{ aws_access_key }}"
+    aws_secret_key: "{{ aws_secret_key }}"
+    region: "{{ aws_region }}"
+    vpc_subnet_id: "{{ aws_subnet }}"
+    exact_count: 1
+    count_tag:
+      Name: "{{ env_id }}-ocp-node-{{ item }}"
+    instance_tags:
+      Name: "{{ env_id }}-ocp-node-{{ item }}"
+      user: sre
```
oybed: same as above

roles/deploy-aws-infra/tasks/main.yml
```diff
@@ -0,0 +1,343 @@
...
+      ocp_role: "{{ infranode_tag }}"
+      env_id: "{{ env_id }}"
+  with_sequence:
+    count="{{ aws_num_infra }}"
+  register: instances_infra
+
+- name: Create Compute instances
+  ec2:
+    image: "{{ aws_image_name }}"
+    instance_type: m4.xlarge
```
oybed: same as above

roles/deploy-aws-infra/tasks/main.yml
```diff
@@ -0,0 +1,343 @@
...
+    group: "{{ aws_node_sgroups }}"
+    termination_protection: yes
```
oybed: same as above

roles/deploy-aws-infra/tasks/main.yml
```diff
@@ -0,0 +1,343 @@
...
+    vpc_subnet_id: "{{ aws_subnet }}"
+    exact_count: 1
+    count_tag:
+      Name: "{{ env_id }}-ocp-node-{{ item }}"
+    instance_tags:
+      Name: "{{ env_id }}-ocp-node-{{ item }}"
+      user: sre
+      role: ocp
+      group: nodes_aws
+      node_labels: '{"region": "primary"}'
```
oybed: same as above

roles/deploy-aws-infra/tasks/main.yml
```diff
@@ -0,0 +1,343 @@
...
+      ocp_role: "{{ computenode_tag }}"
+      env_id: "{{ env_id }}"
+  with_sequence:
+    count="{{ aws_num_nodes }}"
+  register: instances_compute
+
+- name: Give some time until the instances are running before attaching storage
+  pause:
+    seconds: 30
```
oybed: Why 30? Is there a check we can do instead of a fixed sleep?

roles/deploy-aws-infra/tasks/main.yml
```diff
@@ -0,0 +1,343 @@
...
+
+- name: Add EBS for docker storage
+  ec2_vol:
+    aws_access_key: "{{ aws_access_key }}"
+    aws_secret_key: "{{ aws_secret_key }}"
+    region: "{{ aws_region }}"
+    delete_on_termination: yes
+    device_name: /dev/xvdb
```
oybed: Let's use a variable and default to `/dev/xvdb`

roles/deploy-aws-infra/tasks/main.yml
```diff
@@ -0,0 +1,343 @@
...
+    instance: "{{ item.instance_ids[0] }}"
+    volume_size: 40
```
oybed: Need to use variable for size

roles/deploy-aws-infra/tasks/main.yml
```diff
@@ -0,0 +1,343 @@
...
+    volume_type: gp2
+    tags:
+      user: sre
```
oybed: same as above

roles/deploy-aws-infra/vars/main.yml
```diff
@@ -0,0 +1,5 @@
+---
```
oybed: Maybe just add these to the `defaults` file instead (together with all the other variables as mentioned in comments above)

roles/manage-aws-infra/tasks/main.yml
```diff
@@ -0,0 +1,113 @@
+---
```
oybed: Can we please split this into multiple files - i.e.: `stop_start_aws.yml`, `terminate_aws.yml`, etc.

README.md
makentenza: Yes, I agree. Let's make it more generic and also remove the --private-key as suggested below

README.md
makentenza: Agree

inventory/sample-ha.aws.example.com.d/inventory/ec2.ini
makentenza: My mistake. As you can see there is a template directory under sample inventory folder as the initial idea was to generate the .ini file using a jinja template, but as this .ini file is being used by the 'ansible-playbook' execution itself there is a 'chicken and egg' problem there, so we need to figure out how to include this (if we finally think it make sense).

inventory/sample-ha.aws.example.com.d/inventory/group_vars/all.yml
makentenza: Sure

inventory/sample-ha.aws.example.com.d/inventory/group_vars/all.yml
makentenza: We could, but I think this could generate some issues/confusions to anyone using this repo as the AMI name changes on every Region. As an example this AMI in eu-west-1 and us-west-1: 

RHEL-7.4_HVM_GA-20170808-x86_64-2-Access2-GP2 - ami-2b9d6c52
RHEL-7.4_HVM_GA-20170808-x86_64-2-Access2-GP2 - ami-20eec540

inventory/sample-ha.aws.example.com.d/inventory/group_vars/all.yml
makentenza: Sure, that will remove the requirement of setting this values in ec2.ini as well.

inventory/sample-ha.aws.example.com.d/inventory/group_vars/all.yml
makentenza: Sure, will do

inventory/sample-ha.aws.example.com.d/inventory/group_vars/all.yml
makentenza: So the idea here is to automatically create everything related to VPC only if 'aws_create_vpc' is 'true'. In other cases, both subnet (will change to subnetS soon) and security groups must be provided.

What is your suggestion here? To create these SGs and not allow the user to select their owns? My point of view is, if the user wants to use existing subnets, he/she will probably want to use existing SGs as he/she has the knowledge to do that.

inventory/sample-ha.aws.example.com.d/inventory/group_vars/all.yml
makentenza: This is initially only required by OSP:
```
#~ grep -r docker_volume_size . 2>/dev/null | grep -v inventory
./playbooks/dns/roles/openshift-ansible-contrib/playbooks/provisioning/openstack/README.md:The `docker_volume_size` is the default Docker volume size the servers will use.
./playbooks/dns/roles/openshift-ansible-contrib/playbooks/provisioning/openstack/README.md:for master) and change its value. `docker_volume_size` must stay defined as it is
./playbooks/dns/roles/openshift-ansible-contrib/playbooks/provisioning/openstack/stack_params.yaml:master_volume_size: "{{ docker_master_volume_size | default(docker_volume_size) }}"
./playbooks/dns/roles/openshift-ansible-contrib/playbooks/provisioning/openstack/stack_params.yaml:infra_volume_size: "{{ docker_infra_volume_size | default(docker_volume_size) }}"
./playbooks/dns/roles/openshift-ansible-contrib/playbooks/provisioning/openstack/stack_params.yaml:node_volume_size: "{{ docker_node_volume_size | default(docker_volume_size) }}"
./playbooks/nagios/roles/openshift-ansible-contrib/playbooks/provisioning/openstack/README.md:The `docker_volume_size` is the default Docker volume size the servers will use.
./playbooks/nagios/roles/openshift-ansible-contrib/playbooks/provisioning/openstack/README.md:for master) and change its value. `docker_volume_size` must stay defined as it is
./playbooks/nagios/roles/openshift-ansible-contrib/playbooks/provisioning/openstack/stack_params.yaml:master_volume_size: "{{ docker_master_volume_size | default(docker_volume_size) }}"
./playbooks/nagios/roles/openshift-ansible-contrib/playbooks/provisioning/openstack/stack_params.yaml:infra_volume_size: "{{ docker_infra_volume_size | default(docker_volume_size) }}"
./playbooks/nagios/roles/openshift-ansible-contrib/playbooks/provisioning/openstack/stack_params.yaml:node_volume_size: "{{ docker_node_volume_size | default(docker_volume_size) }}"
./playbooks/openshift/provision-openstack.yml:    master_volume_size: "{{ docker_volume_size }}"
./playbooks/openshift/provision-openstack.yml:    node_volume_size: "{{ docker_volume_size }}"
./playbooks/openshift/provision-openstack.yml:    infra_volume_size: "{{ docker_volume_size }}"
./playbooks/openshift/roles/openshift-ansible-contrib/playbooks/provisioning/openstack/README.md:The `docker_volume_size` is the default Docker volume size the servers will use.
./playbooks/openshift/roles/openshift-ansible-contrib/playbooks/provisioning/openstack/README.md:for master) and change its value. `docker_volume_size` must stay defined as it is
./playbooks/openshift/roles/openshift-ansible-contrib/playbooks/provisioning/openstack/stack_params.yaml:master_volume_size: "{{ docker_master_volume_size | default(docker_volume_size) }}"
./playbooks/openshift/roles/openshift-ansible-contrib/playbooks/provisioning/openstack/stack_params.yaml:infra_volume_size: "{{ docker_infra_volume_size | default(docker_volume_size) }}"
./playbooks/openshift/roles/openshift-ansible-contrib/playbooks/provisioning/openstack/stack_params.yaml:node_volume_size: "{{ docker_node_volume_size | default(docker_volume_size) }}"
./playbooks/roles/openshift-ansible-contrib/playbooks/provisioning/openstack/README.md:The `docker_volume_size` is the default Docker volume size the servers will use.
./playbooks/roles/openshift-ansible-contrib/playbooks/provisioning/openstack/README.md:for master) and change its value. `docker_volume_size` must stay defined as it is
./playbooks/roles/openshift-ansible-contrib/playbooks/provisioning/openstack/stack_params.yaml:master_volume_size: "{{ docker_master_volume_size | default(docker_volume_size) }}"
./playbooks/roles/openshift-ansible-contrib/playbooks/provisioning/openstack/stack_params.yaml:infra_volume_size: "{{ docker_infra_volume_size | default(docker_volume_size) }}"
./playbooks/roles/openshift-ansible-contrib/playbooks/provisioning/openstack/stack_params.yaml:node_volume_size: "{{ docker_node_volume_size | default(docker_volume_size) }}"
./roles/openshift-ansible-contrib/playbooks/provisioning/openstack/README.md:The `docker_volume_size` is the default Docker volume size the servers will use.
./roles/openshift-ansible-contrib/playbooks/provisioning/openstack/README.md:for master) and change its value. `docker_volume_size` must stay defined as it is
./roles/openshift-ansible-contrib/playbooks/provisioning/openstack/stack_params.yaml:master_volume_size: "{{ docker_master_volume_size | default(docker_volume_size) }}"
./roles/openshift-ansible-contrib/playbooks/provisioning/openstack/stack_params.yaml:infra_volume_size: "{{ docker_infra_volume_size | default(docker_volume_size) }}"
./roles/openshift-ansible-contrib/playbooks/provisioning/openstack/stack_params.yaml:node_volume_size: "{{ docker_node_volume_size | default(docker_volume_size) }}"```

File: inventory/sample-ha.aws.example.com.d/inventory/hosts (hunk 10 above)
Comment (makentenza) [reply inventory/sample-ha.aws.example.com.d/inventory/hosts]: Not sure, will investigate that

File: inventory/sample.aws.example.com.d/inventory/ec2.ini (hunk 11 above)
Comment (makentenza) [reply inventory/sample.aws.example.com.d/inventory/ec2.ini]: Yes, basically HA is an extension of non-HA inventory so the typos will be the same. Will review both

File: roles/deploy-aws-infra/tasks/main.yml (hunk 13 above)
Comment (makentenza) [reply roles/deploy-aws-infra/tasks/main.yml]: Will do

File: roles/deploy-aws-infra/tasks/main.yml (hunk 16 above)
Comment (makentenza) [reply roles/deploy-aws-infra/tasks/main.yml]: ok

File: inventory/sample-ha.aws.example.com.d/inventory/group_vars/all.yml (hunk 9 above)
Comment (oybed) [reply inventory/sample-ha.aws.example.com.d/inventory/group_vars/all.yml]: Actually, as you'll see from a later comment (see further down), there's a hardcoded size of 40 right now - that needs to be replaced with this variable.

File: roles/deploy-aws-infra/tasks/main.yml (hunk 14 above)
Comment (makentenza) [reply roles/deploy-aws-infra/tasks/main.yml]: What is the advantage of doing that? We can't really re-use this task to remove the instance if exist with "state: absent" as there are some not "compatible" parameters there and also the instances have termination protection so that would fail

File: roles/deploy-aws-infra/tasks/main.yml (hunk 17 above)
Comment (makentenza) [reply roles/deploy-aws-infra/tasks/main.yml]: ok

File: roles/deploy-aws-infra/tasks/main.yml (hunk 15 above)
Comment (makentenza) [reply roles/deploy-aws-infra/tasks/main.yml]: ok

File: inventory/sample-ha.aws.example.com.d/inventory/group_vars/all.yml (hunk 5 above)
Comment (oybed) [reply inventory/sample-ha.aws.example.com.d/inventory/group_vars/all.yml]: Main purpose for having an example is to help guide the user a bit more by having an example value that they may "recognize" by looking in the AWS console (even if it's somewhat different). Alternatively to providing an example value, we could use comments (which we need more of in the example inventories anyway)

File: inventory/sample-ha.aws.example.com.d/inventory/group_vars/all.yml (hunk 8 above)
Comment (oybed) [reply inventory/sample-ha.aws.example.com.d/inventory/group_vars/all.yml]: Agreed. My point was to assign default values in the role and hence don't have to require these to be set in the inventory when `aws_create_vpc` is `true`. In many cases I suspect that the users will be ok with using the default values and hence there's no need to require them in the inventory.

File: roles/deploy-aws-infra/tasks/main.yml (hunk 18 above)
Comment (makentenza) [reply roles/deploy-aws-infra/tasks/main.yml]: Ok, but we should use more than one. From the architecture we designed for Labs there were 2 different flavors; one for Masters and Compute Nodes, and another one for Infra Nodes. Should we use 3 so the user could choose a different flavor for Masters, Compute Nodes and Infra Nodes?

File: roles/deploy-aws-infra/tasks/main.yml (hunk 19 above)
Comment (makentenza) [reply roles/deploy-aws-infra/tasks/main.yml]: sure

File: roles/deploy-aws-infra/tasks/main.yml (hunk 20 above)
Comment (makentenza) [reply roles/deploy-aws-infra/tasks/main.yml]: You mean including a tag called 'env_id'?

File: roles/deploy-aws-infra/tasks/main.yml (hunk 21 above)
Comment (makentenza) [reply roles/deploy-aws-infra/tasks/main.yml]: Sure, but right now there is no option to deploy separate Master(s) and etcd(s) instances, so this will be always true.

But this is a really good topic of discussion; AFAIK OCP 3.7 will require Master and etcd roles to be deployed in different instances, so maybe we could anticipate to this make this change even for 3.6. Thoughts?

File: roles/deploy-aws-infra/tasks/main.yml (hunk 22 above)
Comment (makentenza) [reply roles/deploy-aws-infra/tasks/main.yml]: sure

File: roles/deploy-aws-infra/tasks/main.yml (hunk 25 above)
Comment (makentenza) [reply roles/deploy-aws-infra/tasks/main.yml]: ok

File: roles/deploy-aws-infra/tasks/main.yml (hunk 23 above)
Comment (makentenza) [reply roles/deploy-aws-infra/tasks/main.yml]: Sure, already commented above

File: roles/deploy-aws-infra/tasks/main.yml (hunk 24 above)
Comment (makentenza) [reply roles/deploy-aws-infra/tasks/main.yml]: ok

File: roles/deploy-aws-infra/tasks/main.yml (hunk 27 above)
Comment (makentenza) [reply roles/deploy-aws-infra/tasks/main.yml]: ok

File: roles/deploy-aws-infra/tasks/main.yml (hunk 28 above)
Comment (makentenza) [reply roles/deploy-aws-infra/tasks/main.yml]: ok

File: roles/deploy-aws-infra/tasks/main.yml (hunk 26 above)
Comment (makentenza) [reply roles/deploy-aws-infra/tasks/main.yml]: ok

File: roles/deploy-aws-infra/tasks/main.yml (hunk 29 above)
Comment (makentenza) [reply roles/deploy-aws-infra/tasks/main.yml]: ok

File: roles/deploy-aws-infra/tasks/main.yml (hunk 30 above)
Comment (makentenza) [reply roles/deploy-aws-infra/tasks/main.yml]: There are 2 options here:
- Create a loop and check every N seconds until a limit of M
- Use 'wait' option for ec2 module so the task will not continue until the instances are running. This one has an inconvenient: instances are created one by one instead all at the same time

File: roles/deploy-aws-infra/tasks/main.yml (hunk 31 above)
Comment (makentenza) [reply roles/deploy-aws-infra/tasks/main.yml]: As long as we use the same variable for 'docker_storage_block_device' variable on the inventory file we are fine

File: roles/deploy-aws-infra/tasks/main.yml (hunk 32 above)
Comment (makentenza) [reply roles/deploy-aws-infra/tasks/main.yml]: ok

File: roles/deploy-aws-infra/tasks/main.yml (hunk 33 above)
Comment (makentenza) [reply roles/deploy-aws-infra/tasks/main.yml]: ok

File: roles/deploy-aws-infra/vars/main.yml (hunk 34 above)
Comment (makentenza) [reply roles/deploy-aws-infra/vars/main.yml]: Sure

File: roles/manage-aws-infra/tasks/main.yml (hunk 35 above)
Comment (makentenza) [reply roles/manage-aws-infra/tasks/main.yml]: Sure, no problem

**Comment by makentenza** [comment]
@oybed Thanks for looking at this; just added comments to your first review. Will start working on your suggested changes. Please have a look at my comments as some require a bit more discussion.

File: README.md (hunk 1 above)
Comment (makentenza) [reply README.md]: @oybed what about 'openstack_ssh_public_key'? Should we remove it as well and add that to the README's instructions?

Comment (oybed) [reply README.md]: @makentenza yes, that would be good. Although, we'd also need to supply the public key for AWS (and most likely other cloud environments as well), so maybe we just make this parameter more generic, i.e.: `public_ssh_key` and hence it can *always* be supplied. What do you think?

File: roles/deploy-aws-infra/tasks/main.yml (hunk 14 above)
Comment (oybed) [reply roles/deploy-aws-infra/tasks/main.yml]: Not sure if your response was meant for this comment? But basically, I'd like to have the option to specify a different set of security group rules - hence the variable request?

File: roles/deploy-aws-infra/tasks/main.yml (hunk 18 above)
Comment (oybed) [reply roles/deploy-aws-infra/tasks/main.yml]: yes - if you look at the OpenStack implmentation, it supports the following flavors:
`lb_flavor`, `etcd_flavor`, `master_flavor`, `node_flavor`, `infra_flavor`, `dns_flavor`
I would suggest aligning with the same, but fall back on a "default" flavor if not explicitly specified.

File: roles/deploy-aws-infra/tasks/main.yml (hunk 20 above)
Comment (oybed) [reply roles/deploy-aws-infra/tasks/main.yml]: No, the variable/fact called `env_id` - i.e.: something like `"{{ 'masters_' + env_id }}"`

File: roles/deploy-aws-infra/tasks/main.yml (hunk 21 above)
Comment (oybed) [reply roles/deploy-aws-infra/tasks/main.yml]: ok, so maybe it's a follow-up. If so, let's make sure to write an issue to track.

File: roles/deploy-aws-infra/tasks/main.yml (hunk 30 above)
Comment (oybed) [reply roles/deploy-aws-infra/tasks/main.yml]: Ok, yeah, it just seems that 30 seconds may be a potential failure on a day with a slow connection. There are various ways to do the "wait-loop" to check and break out of the loop if a positive response. Maybe consider something like that.

File: roles/deploy-aws-infra/tasks/main.yml (hunk 32 above)
Comment (oybed) [reply roles/deploy-aws-infra/tasks/main.yml]: yes, please

File: roles/deploy-aws-infra/tasks/main.yml (hunk 31 above)
Comment (oybed) [reply roles/deploy-aws-infra/tasks/main.yml]: sounds good to me

File: README.md (hunk 1 above)
Comment (makentenza) [reply README.md]: @oybed in this case we need to supply the private key. I think the public key is something that is only required for the 'openstack-stack' role in openshift-ansible-contrib if I'm not wrong. For other IaaS providers, including OSP, we will need to specify the private key.

File: inventory/sample-ha.aws.example.com.d/inventory/group_vars/all.yml (hunk 5 above)
Comment (makentenza) [reply inventory/sample-ha.aws.example.com.d/inventory/group_vars/all.yml]: Sure, I have added an example and more comments to clarify this and other values in the inventory files

File: roles/deploy-aws-infra/tasks/main.yml (hunk 13 above)
Comment (makentenza) [reply roles/deploy-aws-infra/tasks/main.yml]: @oybed I see a "problem" here. If we let the user decide about this network, we will need to add more complexity to the playbook in order to figure out what are the possible subnets for that network.

As this is just happening when the user delegates the VPC and all the required components creation to the role ('aws_create_vpc: true'), could we just use the values proposed for the AWS Reference Architecture[1]?

[1 ]https://github.com/openshift/openshift-ansible-contrib/blob/master/reference-architecture/aws-ansible/playbooks/vars/main.yaml#L10-L11

File: roles/deploy-aws-infra/tasks/main.yml (hunk 14 above)
Comment (makentenza) [reply roles/deploy-aws-infra/tasks/main.yml]: @oybed  Sorry, I was just able to view the last 4 lines from the commented code. So, I don't really understand the requirement. Why should we provide the possibility to extend/modify these security groups? As this is an OCP requirement we should ensure that the prereqs are followed, and in case the user needs "extra" SG he/she should be able to add those after the Cluster has been deployed. My main concern here is to avoid OCP installation to fail due to a misconfigured firewall rules.

File: roles/deploy-aws-infra/tasks/main.yml (hunk 18 above)
Comment (makentenza) [reply roles/deploy-aws-infra/tasks/main.yml]: Sure

File: roles/deploy-aws-infra/tasks/main.yml (hunk 20 above)
Comment (makentenza) [reply roles/deploy-aws-infra/tasks/main.yml]: @oybed I think I don't get this one. Where do you see a potential name clash here?

File: roles/deploy-aws-infra/tasks/main.yml (hunk 21 above)
Comment (makentenza) [reply roles/deploy-aws-infra/tasks/main.yml]: Sure

File: roles/deploy-aws-infra/tasks/main.yml (hunk 30 above)
Comment (makentenza) [reply roles/deploy-aws-infra/tasks/main.yml]: @oybed I've finally used the 'wait' option as from the tests I've done it looks to be safer and more reliable.

File: roles/deploy-aws-infra/tasks/main.yml (hunk 13 above)
Comment (oybed) [reply roles/deploy-aws-infra/tasks/main.yml]: @makentenza ok - let's do a follow-up PR to "fix it" later.

File: roles/deploy-aws-infra/tasks/main.yml (hunk 14 above)
Comment (oybed) [reply roles/deploy-aws-infra/tasks/main.yml]: @makentenza ok - not worth spending time on now, we can do a follow-up PR.

File: roles/deploy-aws-infra/tasks/main.yml (hunk 20 above)
Comment (oybed) [reply roles/deploy-aws-infra/tasks/main.yml]: If you spin up multiple clusters using the tools, I believe these will clash. - i.e.: recall how you get "everything" back from the ec2.py tool. In your "hosts" file, you currently have entries like the following:
```
[masters:children]
tag_group_masters_aws
```
... this will then grab masters across all clusters that ec2.py returns unless we make these unique. 

Using the `env_id` aligns with how we do it for OpenStack and it ensures uniqueness.

roles/deploy-aws-infra/tasks/main.yml
oybed: ok - hopefully that doesn't all too much wait time to the execution ... we can do a follow-up though to enhance it later on.

roles/deploy-aws-infra/tasks/main.yml
oybed: BTW: this can also be done as a follow-up PR, but we need it ASAP as we'll want to manage multiple clusters pretty much on day 1.

makentenza:
@oybed PR updated after your review

roles/deploy-aws-infra/tasks/main.yml
makentenza: @oybed you are right. What about adding the following to ec2.ini?

instance_filters = tag:env_id=

There is already an entry called 'regions = ' the user needs to modify, so we could add that as well. I think this could be easier than modify all the entries in the 'hosts' file, so we can maintain that file as it is for every single inventory

oybed:
@makentenza I think this is good to merge as-is to move us forward. Let's do follow-up PRs to fine-tune/adjust as we see needed.

PR #117, opened 2017-10-18
Making pv cleanup part of decommissioning a cluster

etsauer:
#### What does this PR do?
Deletes pvs before cluster delete

Provision a cluster with dynamic provisioning enabled. Ensure some pvcs get created and bound. Then delete the cluster with the following:

```
ansible-playbook -i /path/to/inventory casl-ansible/playbooks/openshift/delete-cluster.yml
```

Now log into cloud and make sure that the kubernetes volumes are all cleaned up.

n/a

cc: @redhat-cop/casl

### DIFF
--- a/playbooks/openshift/delete-cluster.yml
+++ b/playbooks/openshift/delete-cluster.yml
@@ -1,8 +1,8 @@
 ---
 - hosts: seed-hosts
   roles:
-  - roles/openshift-ansible-contrib/openshift-pv-cleanup
+  - openshift-ansible-contrib/roles/openshift-pv-cleanup
 
 - include: delete-openstack.yml
-  when: 
+  when:
   - hosting_infrastructure == 'openstack'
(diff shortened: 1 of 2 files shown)


playbooks/openshift/delete-cluster.yml
```diff
@@ -1,4 +1,7 @@
 ---
+- hosts: seed-hosts
+  roles:
+  - openshift-pv-cleanup
```
oybed: This needs to be referenced from `openshift-ansible-contrib` - i.e.: add the `roles/openshift-ansible-contrib` to the path

playbooks/openshift/delete-cluster.yml
```diff
@@ -1,4 +1,7 @@
 ---
+- hosts: seed-hosts
+  roles:
+  - roles/openshift-ansible-contrib/openshift-pv-cleanup
```
oybed: Actually, no `roles` - see this one for an example:
https://github.com/redhat-cop/casl-ansible/blob/master/playbooks/openshift/delete-openstack.yml#L7

etsauer:
@oybed sorry. I actually tested this time :)

oybed:
Looks good - thank you for submitting this change @etsauer

PR #124, opened 2017-11-15
Inventory vars to set root size for AWS instances

oybed:
#### What does this PR do?
This PR adds the capability of setting the root (/) partition size for instances in AWS as the default 10G is too small. Also, this does some other clean-up for the AWS provisioning - including a bug for the node flavor sizing.

Run the AWS provisioning 

N/A

cc: @redhat-cop/casl


oybed:
@makentenza done - PTAL

makentenza:
Thanks! LGTM. Merging

PR #125, opened 2017-11-15
Added patched ansible container

InfoSec812:
#### What does this PR do?
Adds a Dockerfile which can be used to build a container which includes a patched version of 
ansible which solves an issue with openshift applier

Talk to Øystein!

cc: @oybed


oybed:
LGTM - thank you @InfoSec812

PR #129, opened 2017-11-21
Fix missing dependency for create_users

oybed:
#### What does this PR do?
Adding users with the `htpasswd` module in Ansible fails with missing dependency on python's `passlib`. This PR ensures the dependency is installed.

Define an inventory using htpasswd for an OpenShift cluster, add a set of users per the example inventory, then run the provisioning (alternatively run the `post-install` playbook to call the `create_users` role)

N/A

cc: @redhat-cop/casl


oybed:
@etsauer ping

PR #130, opened 2017-12-01
Add cinder registry volume creation to Openstack install

day4skiing:
#### What does this PR do?

add the following block to `all.yml`
and run `end-to-end.yml` playbook
```
openshift_hosted_registry_storage_kind: openstack
openshift_hosted_registry_storage_access_modes: ['ReadWriteOnce']
openshift_hosted_registry_storage_openstack_filesystem: ext4
openshift_hosted_registry_storage_volume_size: 50Gi

osp_volumes:
- name: "registry-vol"
  display_description: "Registry Volume"
  display_name: "registry-vol"
  size: 50
  purpose: "ose_registry"
```

No
cc: @redhat-cop/casl


day4skiing:
PTAL

inventory/sample.osp.example.com.d/inventory/group_vars/all.yml
```diff
@@ -90,3 +90,17 @@ rhsm_password: ''
 #rhsm_password: ''
 # leave commented out if you want to `--auto-attach` a pool
 #rhsm_pool: ''
+
+# NOTE: these need to be declared in this file because openshift_hosted_registry_storage_kind
+# is needed prior to the creation of OSEv3 hosts. The remainder are here for grouping purposes
+openshift_hosted_registry_storage_kind: openstack
+openshift_hosted_registry_storage_access_modes: ['ReadWriteOnce']
+openshift_hosted_registry_storage_openstack_filesystem: ext4
+openshift_hosted_registry_storage_volume_size: 50Gi
```
oybed: The `volume_size` should be set based on the `osp_volumes` values (and hence removed from here)

playbooks/openshift/cinder-registry.yml
```diff
@@ -0,0 +1,19 @@
...
+# Create a volume for the registry
+- hosts: localhost
+  roles: 
+  - infra-ansible/roles/osp/admin-volume
+
+- hosts: OSEv3
+  tasks:
+  - name: "set registry volumeID fact"
+    set_fact: 
+      openshift_hosted_registry_storage_openstack_volumeID: "{{ disk.volume.id }}"
```
oybed: Per comment above - just add `openshift_hosted_registry_storage_volume_size: "{{ disk.size }}Gi"` here

oybed: ... there may be a better fact to use than `disk.volume.id`, but you probably have the capture so you can decide which one it is.

day4skiing: Yup, `disk.volume.size`

oybed: Thx. Code change looks good. Still need to update the doc and example inventory. Could be good to call-out that these two facts will be automatically set.

day4skiing: Docs updated -- PTAL

Issue #143, opened 2018-01-23
openshift-applier should support a "delete" flag as a way to undo what was done.


etsauer:
Firstly, we could add support for `file_action: delete` in all cases.

Secondly, I would like to be able to have a way to globally set `file_action: delete` on the entire role for a run, as a way to clean up a previous run.

oybed:
PR #152 merged  - closing

PR #152, opened 2018-01-23
Adding ability to delete an openshift-applier run

etsauer:
#### What does this PR do?
Adds ability to delete items in an applier run.

Two modes:

Test delete of entire inventory:
```
ansible-playbook -i inventory/ playbooks/openshift-cluster-seed.yml -e "provision=False"
```

Testing delete of individual items:
```
  - name: "my-cakephp-spaces"
    file: "{{ inventory_dir }}/../files/cakephp/ProjectRequest/my-cakephp-spaces.yml"
    file_action: delete
  - name: "projectrequest-test1"
    template: "{{ inventory_dir }}/../files/ProjectRequest/template.yml"
    params: "{{ inventory_dir }}/../files/ProjectRequest/test1.param"
    template_action: delete
```

https://github.com/redhat-cop/casl-ansible/issues/143

cc: @redhat-cop/casl

### DIFF
--- a/roles/openshift-applier/README.md
+++ b/roles/openshift-applier/README.md
@@ -43,10 +43,12 @@ The `tags` definition is a list of tags that will be processed if the `filter_ta
 
 #### Sourcing a directory with files
 
-You can source a directory composed of static files (without parameters) using `content_dir` instead of defining each file individually. That would look like this:
+You can source a directory composed of static files (without parameters) using `files` instead of defining each file individually. That would look like this:
 ```yaml
 - object: policy
-  content_dir: <dir_with_policy_files>
+  content:
+  - name: directory of files
+    file: <path-to>/directory/
 ```
 In this example above, all of the files in the `<dir_with_policy_files>` directory would get sourced and applied to the cluster (native OpenShift processing).
 
@@ -119,15 +121,18 @@ openshift_cluster_content:
 
 ### Filtering content based on tags
 
-The `openshift-applier` supports the use of tags in the inventory (see example above) to allow for filtering which content should be processed and not. The `filter_tags` variable/fact takes a comma separated list of tags that will be processed and only content/content_dir with matching tags will be applied. 
+The `openshift-applier` supports the use of tags in the inventory (see example above) to allow for filtering which content should be processed and not. The `filter_tags` variable/fact takes a comma separated list of tags that will be processed and only content/content_dir with matching tags will be applied.
 
 **_NOTE:_** Entries in the inventory without tags will not be procssed when a valid list is supplied with the `filter_tags` option.
 
 ```
 filter_tags=tag1,tag2
 
-``` 
+```
 
+### Deprovisioning
+
+The `openshift-applier` role also supports deprovisioning of resources using `file_action: delete` or `template_action: delete`. This can be done either on individual files, if you want to continually ensure objects remain deleted, or it may be done globally using `provision: false`. Using `provision: false` essentially acts like a big 'undo' button, re-running all files and templates through `oc delete -f <resources>`. This can be useful when you want to do a full cleanup to ensure the integrity of you IaC repo, or for simple cleanup while testing changes.
 
 Dependencies
 ------------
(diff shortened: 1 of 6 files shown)


roles/openshift-applier/tasks/process-content.yml
```diff
@@ -4,10 +4,3 @@
   include_tasks: process-one-entry.yml
   with_items:
   - "{{ entry.content | default([]) }}"
-
```
oybed: Can we please deprecate rather than cold-turkey remove it. I.e.: Keep this code, but add a *big fat warning* message about this going away in a near future release.

roles/openshift-applier/tasks/process-one-entry.yml
```diff
@@ -76,6 +83,7 @@
   failed_when:
   - command_result.rc != 0
   - "'AlreadyExists' not in command_result.stderr"
+  - "'NotFound' not in command_result.stderr"
```
oybed: Any chance we get a `NotFound` when in `create` or `apply` mode? If so, this will now be silently ignored ...

roles/openshift-applier/README.md
```diff
@@ -119,15 +121,18 @@ openshift_cluster_content:
...
 
 ```
 filter_tags=tag1,tag2
 
-``` 
+```
+
+### Deprovisioning
 
+The `openshift-applier` role also supports deprovisioning of resources using `file_action: delete` or `template_action: delete`. This can be done either on individual files, if you want to continually ensure objects remain deleted, or it may be done globally using `provision: false`. Using `provision: false` essentially acts like a big 'undo' button, re-running all files and templates through `oc delete -f <resources>`. This can be useful when you want to do a full cleanup to ensure the integrity of you IaC repo, or for simple cleanup while testing changes.
```
oybed: I'd like to see the `provision: false` be called out in a new paragraph as it seems like a major feature. Including an example on how to run it would also be great.

roles/openshift-applier/README.md
```diff
@@ -43,10 +43,12 @@ The `tags` definition is a list of tags that will be processed if the `filter_ta
 
 #### Sourcing a directory with files
 
-You can source a directory composed of static files (without parameters) using `content_dir` instead of defining each file individually. That would look like this:
+You can source a directory composed of static files (without parameters) using `files` instead of defining each file individually. That would look like this:
```
oybed: Can we add a `Note:` below here about `files` replaces `content_dir`, which used to exist ...

roles/openshift-applier/tasks/process-one-entry.yml
etsauer: @oybed I believe the only other cases where you could get a `Not Found` would be if you happened to pass `get` or `replace` as a file_action, which we don't document as valid values.

roles/openshift-applier/tasks/process-content.yml
oybed: @etsauer I'd do something with the `block` feature for the deprecated message, maybe something like the following (just update the message a bit):

```
- name: "Apply OpenShift Cluster Content - based on directory"
  block:
    - fail:
        msg: "This is deprecated ..."
      ignore_errors: True
    - command: >
        oc apply -f {{ tmp_inv_dir }}{{ entry.content_dir }}
  when:
  - entry.content_dir is defined 
  - entry.content_dir|trim != ''
```

roles/openshift-applier/tasks/process-content.yml
```diff
@@ -6,8 +6,11 @@
   - "{{ entry.content | default([]) }}"
 
 - name: "Apply OpenShift Cluster Content ('{{ entry.object }}') - based on directory"
-  command: >
-    oc apply -f {{ tmp_inv_dir }}{{ entry.content_dir }}
+  block:
+  - fail:
+      msg: "This functionality is being deprecated. Please use the 'content.file' variable instead. See role docs for more info."
```
oybed: @etsauer unless you include `ignore_errors: True` here, it will fail ... we probably don't want to fail. Also, could we start the message with an all-caps `WARNING: This functionality ....`

oybed:
@etsauer your branch is out of sync with upstream master, so the dockercloud checks will fail. I'll merge anyway, as the area you are touching doesn't get leveraged as part of the image builds/checks.

PR #166, opened 2018-02-03
Add requirements.yml file to support Galaxy from Ansible Tower

makentenza:
#### What does this PR do?
Add the required file 'requirements.yml' under roles directory to support Galaxy to pull required roles from Tower automatically on project update.

http://docs.ansible.com/ansible-tower/latest/html/userguide/projects.html#ansible-galaxy-support

Create a project in Tower using CASL repository and check roles under 'requirements.yml' are pulled after project update.

No
cc: @redhat-cop/casl

### DIFF
--- a/casl-requirements.yml
+++ b/casl-requirements.yml
@@ -1,6 +1,6 @@
-
 # This is the Ansible Galaxy requirements file to pull in the correct roles
 # to support the operation of CASL provisioning/runs.
+# and also support any role dependency while using the repository in Ansible Tower
 
 # From 'infra-ansible'
 - src: https://github.com/redhat-cop/infra-ansible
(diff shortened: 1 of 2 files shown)


oybed:
@makentenza we already have `casl-requirements.yml` at the top level. I know Tower requires the file to be named `requirements.yml` and exist in the `roles` directory (just like you are adding here). However, I'd prefer to not maintain two of these files, especially as the content seems similar. Here's my suggestion, in priority order:
1. Create a symlink in the `roles` directory to the top level `casl-requirements.yml` - i.e.: `roles/reqiurements.yml --> ../casl-requirements.yml`"

2. If Tower doesn't handle symlinks in the roles directory, move the `casl-requirements.yml` to the roles directory and symlink at the top level -i.e.: `casl-requirements.yml --> roles/requirements.yml`

makentenza:
@oybed Just tested first option and Tower supports it, so let's go that path.

PR #167, opened 2018-02-06
Previous ignore_errors statement didn't work; fixing.

etsauer:
#### What does this PR do?
Fixed previous PR

Run applier with a content_dir defined

n/a

cc: @redhat-cop/casl

### DIFF
--- a/roles/openshift-applier/tasks/process-content.yml
+++ b/roles/openshift-applier/tasks/process-content.yml
@@ -18,7 +18,6 @@
   block:
   - fail:
       msg: "WARNING: This functionality is being deprecated. Please use the 'content.file' variable instead. See role docs for more info."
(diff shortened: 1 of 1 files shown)


roles/openshift-applier/tasks/process-content.yml
```diff
@@ -9,9 +9,9 @@
   block:
   - fail:
       msg: "WARNING: This functionality is being deprecated. Please use the 'content.file' variable instead. See role docs for more info."
-      ignore_errors: True
```
oybed: @etsauer I believe it's in the right place, it just needs to be moved back out to align with the `fail` task. I.e.: we want to ignore errors on the `fail` task, but not on the `command` task. Your propose change will ignore errors on the entire block

Issue #162, opened 2018-01-29
Provision fails when not running `end-to-end` playbook


makentenza:
The following entries in `all.yml` causes the Cluster deployment to fail when running the `config.yml` playbook. It works while running the `end-to-end.yml` playbook.

```
openshift_hosted_registry_storage_kind: openstack
openshift_hosted_registry_storage_access_modes: ['ReadWriteOnce']
openshift_hosted_registry_storage_openstack_filesystem: ext4
```

Deployment ends with the following error:

```
Play:     Create Hosted Resources - persistent volumes
     Task:     Create temp directory for volume definitions
     Message:  The conditional check 'persistent_volumes | length > 0 or persistent_volume_claims | length > 0' failed. The error was: '{{ hostvars[groups.oo_first_master.0] | oo_persistent_volumes(groups) }}: volumeID'
               
               The error appears to have been in '{PATH}/casl-ansible/roles/openshift-ansible/roles/openshift_persistent_volumes/tasks/main.yml': line 2, column 3, but may
               be elsewhere in the file depending on the exact syntax problem.
               
               The offending line appears to be:
               
               ---
               - name: Create temp directory for volume definitions
                 ^ here
```

oybed:
@day4skiing can you please propose a solution for this?

day4skiing:
@oybed Move these values out of all.yml and push the setting of them into the cider-registry.yml  "set registry volumeID fact" play (along with the other facts).  Then trigger off a unique variable to execute the cinder-registry.yml playbook

day4skiing:
@makentenza  @oybed With the changes in #168 the values 
```openshift_hosted_registry_storage_access_modes: ['ReadWriteOnce']
openshift_hosted_registry_storage_openstack_filesystem: ext4```
Are no longer part of `all.yml` .   I tested this last night with the latest version of CASL and the `byo/config.yml` completed cleanly with the new inventory.

PR #168, opened 2018-02-08
fix dup registry on OSP

### LINKED ISSUE #164: openstack provisioning, using cinder for registry volume, should work with multiple clusters
When provisioning multiple clusters within the same tenant, the volume names used to back the registry should be unique (i.e.: `env_id` and `dns_domain`)

day4skiing:
#### What does this PR do?
This PR automatically generates a Unique name per cluster when automatic provisioning on OSP is requested

Define the following in `all.yml` inventory and run `end-to-end.yml`
```
# enable automatic provisioning of cinder registry volume on OSP
casl_storage_registry_kind: 'cinder'
casl_storage_registry_create: True
```
A registry volume defaulted to 20G will be created with the following name `{{ env_id }}.{{ dns_domain }}-registry` which must be unique per tenant.

resolves #164 

cc: @redhat-cop/casl @oybed


docs/PROVISIONING_OPENSTACK.md
```diff
@@ -53,18 +53,10 @@ The following is just an example on how the `sample.osp.example.com` inventory c
...
-osp_volumes:
-- name: "registry-vol"
-  display_description: "Registry Volume"
-  display_name: "registry-vol"
-  size: 50
-  purpose: "ose_registry"
-  ```
-> **Note 1:** `openshift_hosted_registry_storage_volume_size` and `openshift_hosted_registry_storage_openstack_volumeID` are automatically set during play execution
+casl_storage_registry_kind: 'cinder'
+casl_storage_registry_create: True
```
day4skiing: Not sure this flag is necessary

oybed: @day4skiing you really don't need any of these ... what you had for this was just fine

inventory/sample.osp.example.com.d/inventory/group_vars/OSEv3.yml
```diff
@@ -56,6 +56,15 @@ openshift_metrics_cassandra_storage_type: dynamic
 openshift_metrics_cassandra_pvc_size: 40G
 openshift_metrics_duration: 2
 
+# Cinder registry storage
```
oybed: keep these in `all.yml` as otherwise the post-provision-openstack tools won't find them

playbooks/openshift/post-provision-openstack.yml
```diff
@@ -18,7 +18,6 @@
 # provision cinder volume
 - import_playbook: cinder-registry.yml
   when:
-  - openshift_hosted_registry_storage_kind is defined
```
oybed: what you had here worked just fine - not sure why the change? I'd say keep it as-is and just key off the `openshift_hosted_registry_storage_kind` and `openshift_hosted_registry_storage_openstack_volumeID` - i.e.: if the latter is NOT set, we'll provision one to be used ... again, that's the way you had it and should work fine as-is

inventory/sample.osp.example.com.d/inventory/group_vars/all.yml
```diff
@@ -93,18 +93,7 @@ rhsm_password: ''
...
-osp_volumes:
-- name: "registry-vol"
-  display_description: "Registry Volume"
-  display_name: "registry-vol"
-  size: 50
-  purpose: "ose_registry"
+# enable automatic provisioning of cinder registry volume on OSP
+casl_storage_registry_kind: 'cinder'
+casl_storage_registry_create: True
+#casl_registry_size: 20
```
oybed: you can use the `openshift_hosted_registry_storage_volume_size` inventory variable instead of making a `casl` specific one

playbooks/openshift/cinder-registry.yml
```diff
@@ -1,20 +1,28 @@
 --- 
 # Create a volume for the registry
 - hosts: localhost
-  roles: 
+  vars:
+    osp_volumes:
+    - name: "{{ env_id }}.{{ dns_domain }}-registry"
```
oybed: consider using `full_dns_domain` here (and below) - as seen [here](https://github.com/redhat-cop/casl-ansible/blob/master/playbooks/openshift/pre-tasks.yml#L25)

playbooks/openshift/cinder-registry.yml
```diff
@@ -1,20 +1,28 @@
...
+  roles:
   - infra-ansible/roles/osp/admin-volume
 
 - hosts: OSEv3
   tasks:
-  - name: "set registry volumeID fact"
-    set_fact: 
+  - name: "set registry fact"
+    set_fact:
+      openshift_hosted_registry_storage_kind: openstack
```
oybed: this should already be set in the inventory, so no need to set it here

playbooks/openshift/cinder-registry.yml
```diff
@@ -1,20 +1,28 @@
...
       openshift_hosted_registry_storage_openstack_volumeID: "{{ disk.volume.id }}"
       openshift_hosted_registry_storage_volume_size: "{{ disk.volume.size }}Gi"
-    with_items: 
+      openshift_hosted_registry_storage_access_modes: ['ReadWriteOnce']
```
oybed: Let's only set these two if they are not already set...

inventory/sample.osp.example.com.d/inventory/group_vars/OSEv3.yml
day4skiing: If we keep these in `all.yml` I believe it conflicts with the implementation in [openshift-ansible-contrib](https://github.com/openshift/openshift-ansible-contrib/blob/master/playbooks/provisioning/openstack/advanced-configuration.md#use-an-existing-cinder-volume-for-the-openshift-registry)  causing the issue described in  #162

oybed: @day4skiing no, there's no conflict with `contrib` for this. Basically the `post` step doesn't have the necessary values to proceed and hence the individual playbooks cannot be used. Anyway - let's address that in a separate PR to avoid too much cross referencing.

day4skiing:
PTAL

oybed:
@day4skiing I added the `resolves` keyword to the description to auto-close out the issue

docs/PROVISIONING_OPENSTACK.md
```diff
@@ -46,17 +46,7 @@ The following is just an example on how the `sample.osp.example.com` inventory c
 2) Edit `~/src/casl-ansible/inventory/sample.osp.example.com.d/inventory/group_vars/all.yml` to match your cloud provider/environment/project/tenant. See comments in the file for more detailed information on how to fill these in. To utilize OSP cinder storage for the registry add the following to `all.yml`
 ```
 openshift_hosted_registry_storage_kind: openstack
-openshift_hosted_registry_storage_access_modes: ['ReadWriteOnce']
```
oybed: Please include `openshift_hosted_registry_storage_volume_size` here and explain that it can be used to set a proper size. (maybe with an indication that `20` is the default)

inventory/sample.osp.example.com.d/inventory/group_vars/all.yml
```diff
@@ -93,18 +93,14 @@ rhsm_password: ''
...
-
-osp_volumes:
-- name: "registry-vol"
-  display_description: "Registry Volume"
-  display_name: "registry-vol"
-  size: 50
-  purpose: "ose_registry"
+#openshift_hosted_registry_storage_volume_size: 20
+
+# User provisioned Cinder registry storage
```
oybed: Please elaborate to include info indicating that the below will leverage an *existing* Cinder volume with the provided VolumeID.

oybed:
@day4skiing much better - just a few documentation updates and we can merge.

PR #170, opened 2018-02-12
Adding a WIP role for managing Azure infrastructure for OpenShift usi…

logandonley:
Adding a WIP role for managing Azure infrastructure for OpenShift using resource manager templates.

An important thing to note is the template in the templates directory is not being directly used, but the same contents (translated to yaml) are used in the generate-templates.yml parameters for the 'Create Azure Deploy' task. The template will ideally be referenced directly from the Github url once it is complete.

cc: @oybed


oybed:
@logandonley great - thank you. Merging so we can use it as a working point for this feature branch.

PR #197, opened 2018-02-23
aws-provisioning: add new play to create a CNAME for Master API

### LINKED ISSUE #196: aws-provisioning: OCP public API url not accesible after deployment
Route53 entry for public OCP API is not created when `openshift_master_cluster_hostname` and `openshift_master_cluster_public_hostname` have different values:

https://github.com/redhat-cop/casl-ansible/blob/master/inventory/sample.aws.example.com.d/inventory/group_vars/OSEv3.yml#L39-L40

Only the entry to `openshift_master_cluster_hostname` is added:

https://github.com/redhat-cop/casl-ansible/blob/master/roles/manage-aws-infra/tasks/update_dns.yml#L10-L21

We should create a CNAME for this

makentenza:
#### What does this PR do?
Add a new entry in Route53 to cover the `openshift_master_cluster_public_hostname` variable value.

Deploy AWS cluster following instructions from README

resolves #196 

cc: @redhat-cop/casl

### DIFF
--- a/roles/manage-aws-infra/tasks/update_dns.yml
+++ b/roles/manage-aws-infra/tasks/update_dns.yml
@@ -20,6 +20,19 @@
     ttl: 300
   when: not ha_mode
 
+- name: Update Route53 with a CNAME to OCP master API
+  route53:
+    aws_access_key: "{{ aws_access_key }}"
+    aws_secret_key: "{{ aws_secret_key }}"
+    zone: "{{ dns_domain }}."
+    record: "console.{{ env_id }}.{{ dns_domain }}."
+    type: CNAME
+    value: "master-0.{{ env_id }}.{{ dns_domain }}"
+    command: create
+    overwrite: yes
+    ttl: 300
+  when: not ha_mode
+
 - name: Update Route53 with OCP infra wildcard record
   route53:
     aws_access_key: "{{ aws_access_key }}"


roles/manage-aws-infra/tasks/update_dns.yml
```diff
@@ -20,6 +20,19 @@
     ttl: 300
   when: not ha_mode
 
+- name: Update Route53 with a CNAME to OCP master API
+  route53:
+    aws_access_key: "{{ aws_access_key }}"
+    aws_secret_key: "{{ aws_secret_key }}"
+    zone: "{{ dns_domain }}."
+    record: "console.{{ env_id }}.{{ dns_domain }}."
```
etsauer: @makentenza we shouldn't hard code the values this way. Instead, we should allow the user to define `openshift_master_cluster_public_hostname` and `openshift_master_cluster_hostname` and use those names. Note that this will require us to move those variables out to all.yml instead of OSEv3.yml in our sample inventory.

roles/manage-aws-infra/tasks/update_dns.yml
```diff
@@ -20,6 +20,19 @@
...
+    type: CNAME
+    value: "master-0.{{ env_id }}.{{ dns_domain }}"
```
etsauer: same goes here.

oybed:
@makentenza with the new PR #219, I believe you said this one can be closed - correct?

PR #201, opened 2018-03-01
Add deletion of Cinder Registry

### LINKED ISSUE #200: OpenStack provisioned volume stays behind on OCP cluster delete
The automatically provisioned OSP volume for use with OpenShift as a registry PV does not get free'd up when the OCP cluster is deleted. 

The best approach would be to include this PV as part of the Heat stack deployment (as it would automatically be free'd up at delete). For now, we may just need to enhance the `delete-openstack.yml` playbook to free up the volume if it exists.

day4skiing:
#### What does this PR do?
This change will delete the auto-provisioned cinder registry volume when the `delete-cluster` playbook is called

1) run `end-to-end.yml` to provision the cluster on OpenStack with an inventory to auto-provision a registry volume and verify a {{ full_dns_domain }}-registry volume is provisioned in Openstack
2) run `delete-cluster.yml` and verify the volume is deleted upon completion

resolves #200 

cc: @redhat-cop/casl


oybed:
Looks good - thank you.

@day4skiing for future - please add the `resolves` keyword to the issue number to automatically resolve it

PR #204, opened 2018-03-08
Unique galaxy

sherl0cks:
#### What does this PR do?
- supports https://github.com/rht-labs/labs-ci-cd/pull/122
- adds a null check so content like `galaxy_requirements` doesn't blow up the role
- adds regex replace support for `galaxy_requirements` file. needed in the event that this file is self referential.
- move `include` to `include_task` 

I'm currently testing this via labs-ci-cd. https://github.com/sherl0cks/labs-ci-cd/blob/nexus-repos/local-role-test-playbook.yaml

Nope, but @oybed has been working side by side on it

cc: @redhat-cop/casl

### DIFF
--- a/roles/make-applier-projects-unique/tasks/param-file-regex-replace.yml
+++ b/roles/make-applier-projects-unique/tasks/param-file-regex-replace.yml
@@ -4,6 +4,6 @@
     regexp: "{{ pair.regexp }}"
     replace: "{{ pair.replace }}"
   with_items:
-  - "{{ regexp_replace_pairs }} "
+  - "{{ params_regexp_replace_pairs }} "
   loop_control:
     loop_var: pair 
--- a/roles/make-applier-projects-unique/tasks/update-galaxy-requirements.yml
+++ b/roles/make-applier-projects-unique/tasks/update-galaxy-requirements.yml
@@ -1,0 +1,10 @@
+---
+- name: "Set regexp_replace_pairs for galaxy requirements"
+  replace:
+    path: "{{ openshift_cluster_content.galaxy_requirements }}"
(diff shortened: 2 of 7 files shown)


roles/make-applier-projects-unique/tasks/update-galaxy-requirements.yml
```diff
@@ -0,0 +1,10 @@
+---
+- name: "Set regexp_replace_pairs for galaxy requirements"
+  replace:
+    path: "{{ item.galaxy_requirements }}"
+    regexp: "{{ pair.regexp }}"
+    replace: "{{ pair.replace }}"
+  with_items:
+  - "{{ galaxy_regexp_replace_pairs }} "
```
oybed: let's add ` | default([])` in this variable to avoid it bumming out if this variable isn't defined

roles/make-applier-projects-unique/tasks/param-file-regex-replace.yml
```diff
@@ -4,6 +4,6 @@
     regexp: "{{ pair.regexp }}"
     replace: "{{ pair.replace }}"
   with_items:
-  - "{{ regexp_replace_pairs }} "
+  - "{{ params_regexp_replace_pairs }} "
```
oybed: let's add | default([]) in this variable to avoid it bumming out if this variable isn't defined

sherl0cks: hmmm I want to provide the end user a failure message, as these are end user supplied variables. ` | default([])` will cause this to fail silently. perhaps the answer is to provide a `fail`?

PR #205, opened 2018-03-08
Moving OSP client to OSP12 version

oybed:
#### What does this PR do?
This is moving the OSP clients to the "Pike" version which aligns with the OSP12 version.

Perform a provisioning with an OSP 12 environment and ensure it's working. Repeat with an older version of OSP as well.

N/A

cc: @redhat-cop/casl


day4skiing:
LGTM

sabre1041:
Merging per @day4skiing approval

PR #215, opened 2018-03-20
Adding a temporary fix for OCP 3.7 re: docker version

### LINKED ISSUE #213: Repository broken for OCP 3.7 due to docker latest installation
Both 'contrib' and 'infra' repositories are using 'latest' docker, which is `1.13.1` now and the installation for OCP v3.7 will fail with maximum docker version supported error.

oybed:
#### What does this PR do?
Due to incorrect version of docker installed by default (with `latest`), this PR introduces the capability to specify a specific docker version. For 3.9 and later the CASL tools will align with `openshift-ansible` for much of this and hence this will be a `v3.7` workaround only.

**NOTE:** This PR has a dependency on https://github.com/redhat-cop/infra-ansible/pull/150

Specify the `docker_version` in your inventory and run the provisioning. 

resolves #213 

cc: @redhat-cop/casl


makentenza:
@oybed I personally find this workaround 'too complex'. If this is just for 3.7, why don't we just import a new playbook on `end-to-end.yml` when `openshift_release: v3.7` to downgrade docker packages to the supported 1.12 version?

```yum downgrade -y docker-common-1.12.6 docker-1.12.6 docker-client-1.12.6 docker-rhel-push-plugin-1.12.6```

This should work even a new version of docker has been installed and configured.

oybed:
@makentenza the 3.7 workaround is aligned with what's coming in 3.9, so once we moved there, we should be able to just update the playbook and delete this code.

PR #217, opened 2018-03-21
Updating to use infra-ansible for core features, such as rhsm, update-host, etc.

### LINKED ISSUE #35: Subscription-Manager role not idempotent when it comes to enabling repos
If I add or remote yum repos from the list, our role does not recognize that there are changes to be made. We currently do the enabling of repos using commands rather than modules because the modules don't support it yet.

There is an open issue in ansible-core-modules about adding it. https://github.com/ansible/ansible-modules-core/issues/3300

### LINKED ISSUE #160: Variable to deploy metrics deprecated in 3.7
The variable `openshift_hosted_metrics_deploy` can be removed for release 3.7 as is being deprecated in favour of `openshift_metrics_install_metrics` which is already included in sample inventories.

### LINKED ISSUE #216: Update provisioning playbooks to use updated rhsm role instead of subscription-manager role

oybed:
#### What does this PR do?
This PR updates casl-ansible to rely on advanced roles from infra-ansible to take care of what are typically considered "prep" steps for installing OpenShift. This includes:
 - `rhsm` for subscription
 - `config-packages` to install required RPMs
 - `config-selinux` to ensure SELinux is enabled
 - `update-host` to ensure the host is updated (and rebooted if applicable)

**NOTE:** This PR depends on https://github.com/redhat-cop/infra-ansible/pull/154

Perform an installation of a OpenShift cluster using the `end-to-end` playbook

resolves #35 
resolves #216 
resolves #160 

### Other Relevant info, PRs, etc.
 https://github.com/redhat-cop/infra-ansible/pull/154

cc: @redhat-cop/casl


oybed:
@etsauer @makentenza @logandonley there's one use case I'd like to discuss - maybe we can cover this as part of the CASL call:

When re-running the `end-to-end` playbook (or just the `pre-install` playbook) after initial install (i.e.: for updated cluster config or cluster scaling), the `update-host` role will potentially reboot the hosts if there are updates. This is most likely not desirable for operational clusters, yet, it should happen on fresh installs. We may need to find a solution for this - i.e.: potentially by enhancing the `provision-instances` play to determine if the run is the first or for an existing cluster.

oybed:
@etsauer @makentenza @logandonley re: previous comment about host reboots during consecutive runs - had a conversation with @etsauer and we decided to just add a "warning" re: host reboots. Please see updated inventories:
https://github.com/redhat-cop/casl-ansible/pull/217/commits/0f87eafa3975bc1190bab3258a98441ca657457c

PR #218, opened 2018-03-21
Updating dynamic inventory scripts

### LINKED ISSUE #209: should we use `ansible_ssh_host` or `ansible_host` when refreshing dynamic inventory?
Openstack uses one: https://github.com/raffaelespazzoli/casl-ansible/blob/gcp/playbooks/openshift/openstack/provision.yml#L42

Aws uses the other. https://github.com/raffaelespazzoli/casl-ansible/blob/gcp/playbooks/openshift/aws/provision.yml

Looks like our GCP dynamic inventory script only has `ansible_ssh_host` available so i'm guessing thats what we should standardize on.

oybed:
#### What does this PR do?
Create a new top level `files` directory that will hold the *one-and-only* copy of the dynamic inventory scripts. This PR also updates the `openstack.py` script to ensure it has the latest + sets the correct `ansible_*` variables (i.e.: `ansible_hostname`).

This PR also moves the `ec2.py` to the same `files` directory to have a single location for dynamic inventory scripts. 

Perform a provisioning with the new dynamic inventory script part of *your* inventory.

resolves #209 

cc: @redhat-cop/casl


oybed:
@etsauer sure - done

PR #242, opened 2018-04-23
GCE_* Environment variables

victorock:
Use of lookup for environment variables, instead of putting sensitive information in the file.

#### What does this PR do?
Lookup instead of variables in the file.

cc: @redhat-cop/casl


oybed:
No activity in PR for a long time. Closing out due to inactivity. Please re-open if needed.

PR #265, opened 2018-06-27
Add required port for Prometheus to sample inventories

### LINKED ISSUE #263: Add required ports on IPTables for Prometheus and HAProxy stats
As Prometheus is coming GA and will be the default monitoring solution for OCP, we need to prepare the environment to open the required ports

makentenza:
#### What does this PR do?
Add the required ports that need to be opened on the OS side to support Prometheus monitoring using Node Exporter

Follow instructions from README, deploy Prometheus and Node Exporter and check Node Exporter targets and HAProxy are working

Resolves #263 
cc: @redhat-cop/casl

### DIFF
--- a/inventory/sample.gcp.example.com.d/inventory/group_vars/nodes.yaml
+++ b/inventory/sample.gcp.example.com.d/inventory/group_vars/nodes.yaml
@@ -3,6 +3,10 @@
 openshift_node_labels: 
   region: primary
 
+openshift_node_open_ports:
+  - service: "prometheus node exporter"
+    port: "9100/tcp"
+
 openshift_node_kubelet_args:
   kube-reserved: 
   - cpu={{ ansible_processor_vcpus * 50 }}m
--- a/inventory/sample.aws.example.com.d/inventory/group_vars/OSEv3.yml
+++ b/inventory/sample.aws.example.com.d/inventory/group_vars/OSEv3.yml
@@ -90,3 +90,10 @@ openshift_node_kubelet_args:
   - '80'
   image-gc-low-threshold:
   - '60'
+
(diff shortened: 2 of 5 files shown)


oybed:
@makentenza even though this takes care of setting up the nodes, etc. on the OCP side, it won't set up the security groups to allow the traffic outside of the cluster. Do we need to open an issue for this to track it?

makentenza:
@oybed yes, we need to add that to every cloud provider as well.

oybed:
@makentenza ok, great. I opened issue https://github.com/redhat-cop/casl-ansible/issues/271. 

Also, since this has been added to every sample inventory but the Google Cloud one, do you mind adding it there too?

makentenza:
@oybed sure, I though I didn't add that to that inventory sample as there was already an entry for that, but just checked the entry was only for the haproxy stats on the Infra nodes. I've added the node-exporter one to the Nodes definition now.

inventory/sample.gcp.example.com.d/inventory/group_vars/nodes.yaml
```diff
@@ -3,6 +3,10 @@
 openshift_node_labels: 
   region: primary
 
+openshift_node_open_ports:
```
oybed: @makentenza since `openshift_node_open_ports` is is now defined both in `nodes.yaml` and `infra_hosts.yml`, only one of them will be applied to the infra nodes. I believe you'll need to add the prometheus port to both this file and the existing dict in the  `infra_hosts.yml` as well.

makentenza: @oybed yes, you are right, unless we create a new fact combining both dicts they won't be merged. Will push the changes now

PR #268, opened 2018-07-08
Remove not required openshift-applier directory

makentenza:
#### What does this PR do?
Remove the old role directory 'openshift-applier' from roles directory. As this role is also included on the requirements for CASL, the existence of the directory is causing Tasks in Ansible Tower to fail as the directory already exists.

Add a new project in Tower based in CASL and create a job template using this project. Once solved the pre-task updating the info from the repository it must not fail

cc: @redhat-cop/casl


makentenza:
@oybed this is around the discussed issue

PR #272, opened 2018-07-20
Moving roles from the 'openshift-applier' to this repo

oybed:
#### What does this PR do?
Supporting roles used by the [openshift-applier](https://github.com/redhat-cop/openshift-applier) are now being moved (back) to this repo as it's a better fit for these *generic* roles. 

The corresponding 'clean-up' is done in the openshift-applier PR https://github.com/redhat-cop/openshift-applier/pull/33

Just moving roles

N/A

cc: @redhat-cop/casl


sabre1041:
LGTM

PR #274, opened 2018-07-25
Prompt user prior to deleting cluster

tylerauerbeck:
#### What does this PR do?
This PR prompts a user to check whether they're sure that they want to delete their OCP cluster. It prompts a user for a simple yes/no and then takes appropriate action.

**NOTE:** Currently this is using `meta: end_play` and taking action based on the response. It appears to be taking the desired action in `openstack` and ends the play immediately (so that you don't see a number of skipped tasks). However it doesn't appear to behave the same way when skipping the `aws` branch. It still skips everything as expected, it just happens to print out a bunch of skipped tasks (from the `manage_aws_infra` role) instead of ending the play immediately like it does on the `openstack` branch. We may want to take a look at this (either now / in the future), but this is a more of a style/preference issue than a technical one (outside of figuring out why the two playbooks aren't acting the same).

`ansible-playbook -i /path/to/your_casl_inventory/ /path/to/casl-ansible/playbooks/openshift/delete_cluster.yml`

cc: @redhat-cop/casl @oybed @logandonley @makentenza

### DIFF
--- a/playbooks/openshift/delete-cluster.yml
+++ b/playbooks/openshift/delete-cluster.yml
@@ -4,6 +4,17 @@
 #  roles:
 #  - ../../galaxy/openshift-ansible-contrib/roles/openshift-pv-cleanup
 
+- hosts: localhost
+  vars_prompt:
+    name: "delete_cluster"
+    prompt: "Are you sure you want to delete this cluster? (yes/no)"
+    private: no
+  pre_tasks:
+    - set_fact:
+        delete_cluster: "{{ delete_cluster }}"
+    - meta: end_play
+      when: not delete_cluster
+
 - import_playbook: openstack/delete.yml
   when:
   - hosting_infrastructure == 'openstack'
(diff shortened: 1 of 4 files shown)


playbooks/openshift/delete-cluster.yml
```diff
@@ -4,6 +4,17 @@
 #  roles:
 #  - ../../galaxy/openshift-ansible-contrib/roles/openshift-pv-cleanup
 
+- hosts: localhost
+  vars_prompt:
+    name: "delete_cluster"
+    prompt: "Are you sure you want to delete this cluster? (yes/no)"
```
oybed: Please add info on which cluster is about to be deleted - i.e.: cluster id or env_id.

tylerauerbeck: Added env_id to delete prompt.

cc/ @rohitmusti @procrypt

oybed:
@tylerauerbeck instead of having the `meta: end_play` in every playbook (since it only affects "current play"), I wonder if we should just use the `fail` module with a proper `msg` set. That would make the implementation "cleaner". We can discuss.

oybed:
@rohitmusti will finish this as part of PR #285, so closing out this PR

PR #275, opened 2018-07-29
Added tag role based on issue273

### LINKED ISSUE #273: Create role to tag images in OpenShift
This is a move of issue https://github.com/redhat-cop/openshift-applier/issues/26 to this repo since the referenced roles have moved here. Here's a copy/paste of the text in the original issue:

> I want to be able to tag images from different namespace and create labels on that new is in my given project.

gmontalvoy:
#### What does this PR do?
Added a role under casl-ansible/roles/openshift-imagetag that tags images based on {{ source_img }} and {{ img_tag }} in the following format
source_img : openshift/python  (or openshift/python:latest) 
img_tag: python:3
it will be tagged using {{ target_namespace }} (or default if its undefined) 

This has been created following instruction given on issue273 and refered issues. Please review carefully.

resolves #273 

cc: @redhat-cop/casl


roles/openshift-imagetag/tasks/main.yml
```diff
@@ -0,0 +1,12 @@
+---
+
+- name: Setting target_namespace 
+  set_fact: 
+    target_namespace: "{{ target_namespace | default('') }}"
+
+- name:  Tagging as {{ img_tag }} on a namespace {{ target_namespace }}
+  command: >
+    oc tag --insecure=true  {{ source_img }} {{ target_namespace }}/{{ img_tag }}
```
oybed: Please remote the `--insecure=true` portion of this command.  Also, the `target_namespace` will have the `-n` option set (if the variable is set), so need to keep that separate from the actual `img_tag` - i.e.: something like `{{ img_tag }} {{ target_namespace }}`.

gmontalvoy:
Changed what you requested.

Thank you @oybed

oybed:
@gmontalvoy the `resolves` keyword needs to be in the text of the description. I've updated this one above so you can see how it should be done.

PR #295, opened 2018-08-22
Added prerequisites playbook to the end-to-end.yml, changed mountpoin…

### LINKED ISSUE #252: Change FS for Docker to support CRI-O
In order to support CRI-O `/var/lib/docker` can't be mounted as an FS, which is now mounted. We need to move this to `/var/lib/containers/docker`

gmontalvoy:
…t from /var/lib/docker to /var/lib/containers/docker and updated example inventories to label all nodes with runtime=cri-o'

#### What does this PR do?
Changes needed to support deployment with cri-o.
- Changed /var/lib/docker to /var/lib/containers/docker
- Included prerequisites.yml from openshift-ansible to end-to-end.yml
- Updated sample inventories so nodes will be labeled as runtime=cri-o

resolves  #252 

cc: @redhat-cop/casl


playbooks/openshift/prep-inventory.yml
```diff
@@ -49,7 +49,7 @@
       docker_data_size: "95%VG"
       docker_dm_basesize: "3G"
       container_root_lv_name: "docker_lv"
-      container_root_lv_mount_path: "/var/lib/docker"
+      container_root_lv_mount_path: "/var/lib/containers/docker"
```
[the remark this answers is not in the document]
gmontalvoy: Ok, I will refactor as a variable inside group_vars at each inventory.

makentenza:
@oybed you are right, is a requirement to add an additional TCP port to your external firewall in order to enable both `oc exec` and `oc rsh` 

```
If using CRIO, you must open tcp/10010 to allow oc exec and oc rsh operations.
```

Regarding labelling, the requirement is not to label Nodes with an specific label for Docker GC after Builds, the requirement is to match existing Nodes labels where Builds are performed with the variable `openshift_crio_docker_gc_node_selector`.

This is a simple thing, GC will run as a DaemonSet on every single Node where it's selector matches the Node Selector, so what to put `openshift_crio_docker_gc_node_selector` variable is up to the user depending where the specific configuration done on the Cluster for Builds.

If we want to be "safe", we can hardcode both, i.e. add a new custom label to every single Node (like runtime=cri-o from the example) and use that label for `openshift_crio_docker_gc_node_selector` but it could be the case where we could end running these GC Pods where not required; this should not be harmful but is something to consider.

Finally, and just for heads up, with the actual configuration around `container_root_lv_mount_path`, the mountpoint for Docker storage is configured properly...

```
[root@app-node-0 ~]# df -Th | grep -i "containers/docker"
/dev/mapper/docker_vg-docker_lv xfs        20G  1.3G   19G   7% /var/lib/containers/docker

[root@app-node-0 ~]# docker info | grep -i "containers/docker"
Docker Root Dir: /var/lib/containers/docker
``` 

...even the storage configuration file has the expected wrong values

```
[root@app-node-0 ~]# cat /etc/sysconfig/docker-storage-setup | grep -i path
CONTAINER_ROOT_LV_MOUNT_PATH="/var/lib/docker"
```

Didn't have the time to check this further yet, but is at least "interesting"

gmontalvoy:
Thank you for the comments,
I added required ports (10010/TCP) on group_vars app_hosts.yml infra_hosts.yml and masters.yml at each inventory.

@makentenza 
As you can see here: https://github.com/openshift/openshift-ansible/blob/1597613fa0acee2d04a16effed9bebb023acce83/roles/openshift_docker_gc/defaults/main.yml

if no `openshift_crio_docker_gc_node_selector` a runtime=cri-o will be the default value to looking for create gc pods as long as `openshift_crio_enable_docker_gc=true`

as far as it depends on the user configuration where to generate builds for their applications, would you agree if we hardcode only app_nodes as runtime=cri-o ?

About mount /var/lib/containers/docker or /var/lib/docker, I set a variable in playbooks/openshift/end-to-end.yml with some examples on inventories so user can define the mount point.

Any feedback is always appreciated.

Thank you!

oybed:
@gmontalvoy so port 10010 for the actual hosts themselves I believe is handled by `openshift-ansible`, so no need to add that to the inventories like you just did. However, what @makentenza and I were saying before is that this port needs to be opened *externally* - i.e.: in the security groups, etc. For AWS, GCP, etc. this should be pretty straight forward as it's all part of this repo. For OSP, however, it is a bit more complicated as we no longer control that code in this repo but rather getting it from `openshift-ansible` (part of PR #294). I think this is something we need to explore a bit more and maybe it's just handled as a follow-up PR.

inventory/sample.byo.example.com.d/inventory/group_vars/OSEv3.yml
```diff
@@ -51,8 +51,6 @@ openshift_node_open_ports:
     port: "1936/tcp"
   - service: "prometheus node exporter"
     port: "9100/tcp"
-
```
oybed: nit-pick - can you please bring back the blank lines

playbooks/openshift/prep-inventory.yml
```diff
@@ -49,7 +49,7 @@
       docker_data_size: "95%VG"
       docker_dm_basesize: "3G"
       container_root_lv_name: "docker_lv"
-      container_root_lv_mount_path: "/var/lib/docker"
+      container_root_lv_mount_path: "{{ docker_storage_mount_point | default('/var/lib/docker') }}"
```
makentenza: We should default this to `/var/lib/containers/docker` to be aligned with what we are doing in `openshift-ansible` respository

makentenza:
@oybed @gmontalvoy Just a minor change request from my side and we are good to go for merging.

gmontalvoy:
defaulted to /var/lib/containers/docker

makentenza:
Thanks @gmontalvoy 

LGTM

PR #302, opened 2018-08-31
Improvements to setup for AWS docs

etsauer:
#### What does this PR do?

cc: @redhat-cop/casl

### DIFF
--- a/docs/PROVISIONING_AWS.adoc
+++ b/docs/PROVISIONING_AWS.adoc
@@ -1,5 +1,7 @@
 = OpenShift on AWS EC2 using CASL
 
+TODO: Using an existing VPC or maintaining the created VPC is not supported yet.
+
 == Local Setup (one time, only)
 
 NOTE: These steps are a canned set of steps serving as an example, and may be different in your environment.
@@ -8,11 +10,11 @@ Before getting started following this guide, you'll need the following:
 
 * Access to an AWS account with the proper policies to create resources (see details below)
 * Docker installed
-  * RHEL/CentOS: `yum install -y docker`
-  * Fedora: `dnf install -y docker`
-  * **NOTE:** If you plan to run docker as yourself (non-root), your username must be added to the `docker` user group.
+  ** RHEL/CentOS: `yum install -y docker`
+  ** Fedora: `dnf install -y docker`
+  ** **NOTE:** If you plan to run docker as yourself (non-root), your username must be added to the `docker` user group.
 * Ansible 2.5 or later installed
-  * link:https://docs.ansible.com/ansible/latest/installation_guide/intro_installation.html[See Installation Guide]
+  ** link:https://docs.ansible.com/ansible/latest/installation_guide/intro_installation.html[See Installation Guide]
 [source,bash]
 ----
 cd ~/src/
@@ -34,7 +36,7 @@ ansible-galaxy install -r casl-requirements.yml -p galaxy
 The following needs to be configured in your AWS account.
 
 * An link:https://console.aws.amazon.com/iam/home?#/users[IAM User] must be created with API access, and the ability to provision VPCs, Subnets, Instances, EBS Volumes, etc. (see note below on policy specifics)
-  * Once created, set export `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` to your environment.
+  ** Once created, set export `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` to your environment.
 * Create a link:http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-key-pairs.html#having-ec2-create-your-key-pair[Key-pair in AWS]
 * Create a link:https://console.aws.amazon.com/route53/home?#hosted-zones:[Hosted Zone] in your AWS Account.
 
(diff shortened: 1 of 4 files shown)


etsauer:
@oybed @makentenza @logandonley @tylerauerbeck

etsauer:
@makentenza that's a good callout. I wasn't aware there even was a README in the role. Typically we've used the role README to capture supported role variables/parameters, but i think in this case there really aren't any "role variables" like we currently have in openstack. @oybed what are your thoughts?

docs/PROVISIONING_AWS.adoc
```diff
@@ -0,0 +1,142 @@
...
+
+== Local Setup (one time, only)
+
+NOTE: These steps are a canned set of steps serving as an example, and may be different in your environment.
+
+Before getting started following this guide, you'll need the following:
+
+* Access to an AWS account with the proper policies to create resources (see details below)
+* Docker installed
+  * RHEL/CentOS: `yum install -y docker`
```
oybed: These don't show up as sub-bullets in `.adoc` format. Can you please fix?

oybed: Apples to all "Sub-bullets" throughout the doc

docs/PROVISIONING_AWS.adoc
```diff
@@ -0,0 +1,142 @@
...
+cd ~/src/casl-ansible
+ansible-galaxy install -r casl-requirements.yml -p galaxy
+----
+
+== AWS Setup
+
+The following needs to be configured in your AWS account.
+
+* An link:https://console.aws.amazon.com/iam/home?#/users[IAM User] must be created with API access, and the ability to provision VPCs, Subnets, Instances, EBS Volumes, etc. (see note below on policy specifics)
+  * Once created, set export `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` to your environment.
```
oybed: This is only needed if the docker image isn't used, and it's handled by the `installer-aws` image as documented here if running with the CASL provided container: https://github.com/redhat-cop/casl-ansible/tree/master/images/installer-aws ... we should "merge" these docs to ensure it's covered correctly, and in **one** place only.

docs/PROVISIONING_AWS.adoc
```diff
@@ -0,0 +1,142 @@
...
+* Create a link:http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-key-pairs.html#having-ec2-create-your-key-pair[Key-pair in AWS]
+* Create a link:https://console.aws.amazon.com/route53/home?#hosted-zones:[Hosted Zone] in your AWS Account.
```
oybed: This isn't necessarily needed (e.g: many users already have this set up), so we need to ensure we qualify this with the proper text.

docs/PROVISIONING_AWS.adoc
```diff
@@ -0,0 +1,142 @@
...
+
+=== Example Policy Assignment
```
oybed: I think this is "muddying" the waters quite a bit here, and I think it's very specific to a specific user (or just a few users). I'd prefer that we don't include as much as it is here within these docs, but maybe link out to another doc if/when this section is needed?

docs/PROVISIONING_AWS.adoc
```diff
@@ -0,0 +1,142 @@
...
+
+NOTE: Make sure the region matches your target region above and that you insert an ARN that matches your account.
+
+Cool! Now you're ready to provision OpenShift clusters on AWS
+
+== Provision an OpenShift Cluster
+
+As an example, we'll provision the `sample.aws.example.com` cluster defined in the `~/src/casl-ansible/inventory` directory.
+
+NOTE: It is recommended that you use a different inventory similar to the ones found in the `~src/casl-ansible/inventory` directory and keep it elsewhere. This allows you to update/remove/change your casl-ansble source directory without losing your inventory. Also note that it may take some effort to get the inventory just right, hence it is very beneficial to keep it around for future use without having to redo everything.
```
oybed: Maybe reword this to something like:
> Note: Unless you already have a working inventory, it is recommended that you make a copy of the above mentioned sample inventory and keep it somewhere outside of the `casl-ansible` directory. This allows you to ....

docs/PROVISIONING_AWS.adoc
```diff
@@ -0,0 +1,142 @@
...
+
+The following is just an example on how the `sample.aws.example.com` inventory can be used:
```
oybed: Probably just remove this line, or combine it with the one at line 76

docs/PROVISIONING_AWS.adoc
```diff
@@ -0,0 +1,142 @@
...
+
+1. Select an link:https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.RegionsAndAvailabilityZones.html[AWS Region] you would like to provision your cluster in.
+  * Modify the `regions` entry (line 13) in the inventory `ec2.ini` with your region
+  * Modify the `cloud_infrastructure.region` variable in `group_vars/all.yml` with your region.
+2. Select an _environment id_. This is a shortname for your cluster. In the example cluster domain `sample.aws.example.com`, your _environment id_ would be `sample`.
+  * Modify `instance_filters` entry (line 14) in the inventory `ec2.ini` file to match your chosen _environment id_
+  * Set the `env_id` variable in `group_vars/all.yml` to match.
+3. Edit `group_vars/all.yml` to match your AWS environment. See comments in the file for more detailed information on how to fill these in.
+4. Edit `group_vars/OSEv3.yml` for your AWS specific configuration. See comments in the file for more detailed information on how to fill these in.
```
oybed: Probably should reword "... for your AWS specific configuration." to something like "... for OpenShift Container Platform related configuration."

docs/PROVISIONING_AWS.adoc
```diff
@@ -0,0 +1,142 @@
...
+
+Run the `end-to-end` provisioning playbook via our link:../images/installer-aws/[AWS installer container image].
+
+[source,bash]
+----
+docker run -u `id -u` \
+      -v $HOME/.ssh/id_rsa:/opt/app-root/src/.ssh/id_rsa:Z \
+      -v $HOME/src/:/tmp/src:Z \
+      -e AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID \
```
oybed: As mentioned above, the `installer-aws` container takes care of sourcing the AWS keys from the `csv` file downloaded from AWS, so let's make sure we update and align these docs.

docs/PROVISIONING_AWS.adoc
```diff
@@ -0,0 +1,142 @@
...
+      -e AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY \
+      -e INVENTORY_DIR=/tmp/src/casl-ansible/inventory/sample.aws.example.com.d/inventory \
+      -e PLAYBOOK_FILE=/tmp/src/casl-ansible/playbooks/openshift/end-to-end.yml \
+      -e OPTS="-e aws_key_name=my-key-name" -t \
+      redhatcop/installer-aws
+----
+
+NOTE: The `aws_key_name` variable at the end should specify the name of your AWS keypair - as noted under AWS Specific Requirements above.
```
oybed: I know there has been confusion around the API access keys v.s. the AWS keypair. Not sure how we could maybe try to clarify it a bit more, but maybe it's worth calling it out here that this is **not** the API access keys.

docs/PROVISIONING_AWS.adoc
```diff
@@ -0,0 +1,142 @@
...
+
+NOTE: The above bind-mounts will map files and source directories to the correct locations within the control host container. Update the local paths per your environment for a successful run.
```
oybed: It may also be worth nothing that depending on your host used to run the container, it may or may not need the `:Z` on the mounts.

docs/PROVISIONING_AWS.adoc
```diff
@@ -0,0 +1,142 @@
...
+----
+docker run -u `id -u` \
+      -v $HOME/.ssh/id_rsa:/opt/app-root/src/.ssh/id_rsa:Z \
+      -v $HOME/src/:/tmp/src:Z \
+      -e INVENTORY_DIR=/tmp/src/casl-ansible/inventory/sample.casl.example.com.d/inventory \
+      -e PLAYBOOK_FILE=/tmp/src/casl-ansible/playbooks/openshift/delete-cluster.yml \
+      redhatcop/installer-aws
+----
+
+TODO: Using an existing VPC or maintaining the created VPC is not supported yet.
```
oybed: We may want to call this out early on in the docs rather than as just a small note at the end. Also, we should get this supported ASAP as the CASL tools used to support this use-case.

oybed:
@etsauer @makentenza re: role README - there's good information in the role's readme that will be lost as-is if we just remove it. Hence, I'd vote for keeping it as a "more specific" set of details for people that are interested. The challenge (as with all static docs) is that it's easily out of date (just like this README was) and hence it adds more overhead to maintenance. 

There's also the `installer-aws` README (which I commented on inline earlier).

oybed:
@etsauer any updates on this one? I'd like to get it merged.

etsauer:
@oybed addressed some of your comments. I think the others can wait until we complete work on the installer-aws image.

PR #305, opened 2018-09-07
Missing node group name fix

darthlukan:
#### What does this PR do?
`openshift-ansible` 3.10 installer expects masters to have `openshift_node_group_name='node-config-master-infra'` set. This PR includes the relevant changes to the BYO `hosts` and `group_vars/OSEv3.yml` files in order to successfully complete the tasks.

1. Set infrastructure for 3 masters and use release-3.10 of `openshift-ansible`
2. DO NOT include the changes in this PR
3. Run `ansible-playbook -i /path/to/byo/inventory /path/to/byo/playbook.yml`
4. Note the failure with error:
```
... SNIP additional info prior
Missing openshift_node_group_name entry node-config-master-infra
... SNIP additional info after
```
5. Merge changes from this PR
6. Repeat step 3
7. Note that error does not occur.

[This commit](https://github.com/openshift/openshift-ansible/commit/41e4c8dc5b0d47d56b1b7c7e7cbf5dd81b0202c6
) is included in openshift-ansible 3.10.0+ branches as a default fact.
 
cc: @redhat-cop/casl

### DIFF
--- a/inventory/sample.byo.example.com.d/inventory/group_vars/OSEv3.yml
+++ b/inventory/sample.byo.example.com.d/inventory/group_vars/OSEv3.yml
@@ -70,6 +70,15 @@ openshift_node_groups:
     value: [ 'cpu={{ ansible_processor_vcpus * 50 }}m', 'memory={{ ansible_processor_vcpus * 50 }}M' ]
   - key: kubeletArguments.system-reserved
     value: [ 'cpu={{ ansible_processor_vcpus * 50 }}m', 'memory={{ ansible_processor_vcpus * 100 }}M' ]
+- name: node-config-master-infra
+  labels:
+    - 'node-role.kubernetes.io/infra=true'
+    - 'node-role.kubernetes.io/master=true'
+  edits:
+  - key: kubeletArguments.kube-reserved
+    value: [ 'cpu={{ ansible_processor_vcpus * 50 }}m', 'memory={{ ansible_processor_vcpus * 50 }}M' ]
+  - key: kubeletArguments.system-reserved
+    value: [ 'cpu={{ ansible_processor_vcpus * 50 }}m', 'memory={{ ansible_processor_vcpus * 100 }}M' ]
 - name: node-config-compute
   labels:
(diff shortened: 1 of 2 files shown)


oybed:
@darthlukan A couple of concerns here:
1. The commit you are referencing outlines how this should *not* be used for production, but more of an all-in-one or hosts with combined roles
1. The change to the `hosts` file conflicts with [the entry that already exists for the masters](https://github.com/redhat-cop/casl-ansible/blob/master/inventory/sample.byo.example.com.d/inventory/group_vars/masters.yml#L3)

Due to 2. above, we cannot accept this PR as-is, but maybe there's an alternative approach that can be taken?

darthlukan:
@oybed - Fair enough. 

Regarding 1: I understand that it's not recommended for production, but then we have a catch-22 situation. It's not recommended for production use, but it's the (simplest) way to avoid the error for a successful install. Perhaps there's another issue and this one is just a symptom?

Regarding 2: Related to 1, this may be a symptom of another issue and is not needed at all.

For both, I'll seek out additional guidance and see if perhaps this whole PR is unnecessary.

oybed:
@darthlukan is this a fatal error you are seeing? We successfully use the CASL approach with v3.10 as-is, and we're not seeing this error, but unless it's a fatal error we may just not notice it.

darthlukan:
Yes, it's fatal. There's no way for the installation to continue once we hit that play.

darthlukan:
Follow-up: TL;DR: This PR is not needed.

The problems we ran into with the `openshift_node_group_name` that this PR was meant to fix were related to an incorrectly configured `OSEv3.yml` file. Specifically, I did not replace the default  `osm_default_node_selector` value which snowballed into the group issue. Once I was given the primer on that variable and its behavior in OCP 3.10, things worked as expected.

oybed:
@darthlukan thanks for the update. Yes, the `osm_default_node_selector` seems to have bitten many attempts in `v3.10`. We should get the sample inventories in CASL updated to reflect this. 

Anyway, closing this PR as not needed/invalid.

PR #308, opened 2018-09-17
Fix GCP Registry Storage

tylerauerbeck:
#### What does this PR do?
This PR updates CASL to utilize the openshift-ansible OSv3 variables to utilize a GCP cloud bucket as the backend storage for the registry. This allows us to remove some of the custom logic that we had built in (which were causing some issues) and now provides a working backend for the registry.

Create a 1x1x1 cluster and kick off a build. You will now see that the image is successfully pushed to the registry. Previously we would see a 500/403 error reported showing that we didn't have permissions to utilize the GCP storage bucket.

N/A

cc: @redhat-cop/casl


roles/manage-gcp-infra/tasks/cleanup.yml
```diff
@@ -68,6 +68,16 @@
...
   changed_when: false
+
+- name: Check registry bucket for files
+  shell: gsutil ls -l gs://{{ google_registry_bucket_name }}
+  register: gs_file_count
+
+- name: Empty registry bucket contents
+  shell: gsutil rm gs://{{ google_registry_bucket_name }}/**
+  when:
+  - deployment.stdout
```
oybed: Since `deployment` is being checked here, I presume it's for knowing that a deployment exists? If so, probably shave have the same check for the "Check registry bucket for files" above? Preferably just group these in a `block` checking for a valid deployment. Also, is any `stdout` categorized as a deployment exists? ... or is there a better check to be done here? I.e.: I see line 69 relies on a return code - maybe that's better here too?

playbooks/openshift/gcp/post-install.yml
```diff
@@ -2,8 +2,6 @@
 
 - hosts: masters[0]
   tasks:
-  - name: configure registry
-    include_role:
-      name:  manage-gcp-infra
-      tasks_from: configure-registry-object-storage.yml
-
+  - name:"Placeholder for GCP Post-install"
```
oybed: If there are no post-steps, let's just remove the file... or alternatively just lave it "blank" (only the first `---` left in the file)

inventory/sample.gcp.example.com.d/inventory/gce.ini
```diff
@@ -41,9 +41,9 @@ libcloud_secrets =
 
 # If you are not going to use a 'secrets.py' file, you can set the necessary
 # authorization parameters here.
-gce_service_account_email_address = <REPLACE WITH A VALID SERVICE ACCOUNT E-MAIL ADDRESS>
-gce_service_account_pem_file_path = <REPLACE WITH A VALID SERVICE ACCOUNT PEM/JSON FILE>
-gce_project_id = <REPLACE WITH A VALID PROJECT ID>
+gce_service_account_email_address = "{{ lookup('env','GCE_EMAIL') }}"
```
oybed: Curious about using environment variables for this. That means that they could change for a follow-up run. Is that going to be a problem? May want to just keep them as part of the inventory if so.

playbooks/openshift/gcp/post-install.yml
tylerauerbeck: I think I'll just leave the `---` for now. I think this will still be useful to have in the future should we want to do things like enable and configure other gcp components.

roles/manage-gcp-infra/tasks/cleanup.yml
tylerauerbeck: @oybed I've updated this to block the cluster cleanup components together. From what I'm seeing ( and using), any stdout from both `deployment` and `gs_file_count` show that there is a valid deployment and existing files in the storage bucket. 

On another note, it looks like there's an existing module to deal with gcp storage. It just looks like it relies on tokens versus being able to use the json file that we're using now. It may be worth while to see if we can get this to work with the json file or open up an issue with the ansible folks to see if we can improve these modules. I'd much rather be able to lean on modules versus having to rely on the right packages/versions to be installed on a system for this to work.

sabre1041:
@tylerauerbeck are the changes that you committed 30 days ago ready for a rereview?

tylerauerbeck:
@sabre1041 @oybed Sorry for the delay. Did some cleanup on this and this is ready for review.

Issue #266, opened 2018-07-05
Add multi-device support for CNS in AWS


makentenza:
Add multi-device support for CNS in AWS

PR #311, opened 2018-09-18
Fix #310, byo node gluster var fix

### LINKED ISSUE #310: BYO: 'dict object' has no attribute 'cns_node_glusterfs_volume' - OCS 3.10
This error occurs because there's no variable by that name defined in `/path/to/inventory/host_vars/localhost`. Previously, I had that variable in there as such:

```
cns_node_glusterfs_volume:
- "/dev/sdc"
- "/dev/sdd"
```
Which allowed me to not get this error, but triggered the error associated with [this issue](https://github.com/openshift/openshift-ansible/issues/5071), the fragile workaround for which is to remove global definitions of glusterfs devices.

The current behavior of `cns_node_glusterfs_volume` is that it only accepts a single device, which means it is in conflict with `glusterfs_devices` set in the host inventory when multiple devices are specified as a JSON array.

The suggested fix (via @oybed in external conversation) is two pronged:
1. `cns_node_glusterfs_volume` should do as intended and allow an array of devices to be specified.
2. `cns_node_glusterfs_volume` should be ignored in the `pre-install` playbook if `glusterfs_devices` is set in the hosts inventory.

This issue came up during a byo use-case. It may affect other use-cases if they include OCS 3.10.

darthlukan:
#### What does this PR do?
Fixes issue #310 by not requiring `cns_node_glusterfs_volume` when `glusterfs_devices` is already set.

Using byo sample as a base:
1. Do not set `cns_node_glusterfs_volume` in `host_vars/localhost`
2. Add `[glusterfs]` and `[glusterfs_registry]` with appropriate nodes per OCP 3.10 docs
3. Run inventory against byo-infra.yml playbook
4. Note that casl-ansible fails on pre-install.yml (fails to set `glusterfs_devices` fact)

Solution:
1. Merge this PR
2. Change nothing from above failure procedures
3. Note that casl-ansible continues as expected.

#310 

cc: @redhat-cop/casl


oybed:
Will merge as-is and complete the 2nd half of issue #311 as part of issue #266. (i.e.: making `cns_node_glusterfs_volume` as list of devices (even when the list is of length 1).

PR #316, opened 2018-09-27
Updates examples with new mandatory var oreg_url, oreg_auth_user and oreg_auth_password.

jtudelag:
#### What does this PR do?
Adds new mandatory var oreg_url, oreg_auth_user and oreg_auth_password.
Since 3.10 there is a check to validate this vars.
In 3.11, all container images from the new registry.redhat.io will need credentials to be pulled.

cc: @redhat-cop/casl

### DIFF
--- a/inventory/s1-ha.osp.example.com.d/inventory/group_vars/OSEv3.yml
+++ b/inventory/s1-ha.osp.example.com.d/inventory/group_vars/OSEv3.yml
@@ -12,6 +12,12 @@ openshift_master_cluster_hostname: "{{ groups.lb.0 }}"
 openshift_master_cluster_public_hostname: "{{ groups.lb.0 }}"
 openshift_master_default_subdomain: "apps.{{ env_id }}.{{ dns_domain }}"
 
+# Registry URL & Credentials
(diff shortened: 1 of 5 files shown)


jtudelag:
@tylerauerbeck makes sense, changed.

jtudelag:
@tylerauerbeck PTAL ;)

oybed:
@jtudelag Don't think this is "mandatory" anymore. It's a good one to include, but we may want to also comment out the `oreg_url` to make it serve as just a sample portion of the inventory.

tylerauerbeck:
@oybed Actually it looks like this may actually be required now. The most recent cluster I tried to deploy with release 3.10.1 fails unless it has `oreg_auth_user` and `oreg_auth_password` passed in.

oybed:
@tylerauerbeck I've heard mixed info on this - including that it was a bug causing it to be "required" for v3.10, and this bug has now been fixed. The latest `v3.10.x` clusters I deployed didn't require it. Either way, having the info in the sample inventory is fine, but let's do a bit more research to figure out if it we should leave it commented in/out.

PR #320, opened 2018-10-11
Opened up node exporter port on all OCP hosts

sabre1041:
#### What does this PR do?
Opens up node exporter port on all OCP nodes. Needed to support cluster monitoring which is enabled by default in OCP 3.11

Validate configuration to open port is correct

cc: @redhat-cop/casl

### DIFF
--- a/roles/manage-aws-infra/templates/casl-aws-cformation.yml.j2
+++ b/roles/manage-aws-infra/templates/casl-aws-cformation.yml.j2
@@ -247,6 +247,10 @@ Resources:
         FromPort: '4789'
         ToPort: '4789'
         CidrIp: 0.0.0.0/0
+      - IpProtocol: tcp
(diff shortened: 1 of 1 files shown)


oybed:
@sabre1041 the file in the `archive` area isn't used anymore, so no need to change that one (+we should remove that content). The challenge we're having here is to enable this for OSP since we don't control that code. I think we need to work on getting this "pulled together" to become the "upstream" again for OSP. 

Also, we should make the changes for AWS, not just Google Cloud.

@tylerauerbeck no - this is for the actual IaaS. The configuration in the `*_hosts` files are for OpenShift to open the ports - both needed.

sabre1041:
@oybed @tylerauerbeck I updated the PR to remove the modification to the file in the archive directory.

Please review at your convenience

Issue #339, opened 2018-11-21
Add role to be able to apply labels based on regular expression


InfoSec812:
It would be useful to have a role which applies labels to OpenShift objects based on the object name matching a given regular expression.

PR #340, opened 2018-11-21
Adds a new role (openshift-label-by-expression) which allows you to apply a label to all objects (except Pods) which match a given expression

InfoSec812:
#### What does this PR do?
Adds a new role (openshift-label-by-expression) which allows you to apply a label to all objects (except Pods) which match a given expression.

Apply the following example playbook:
```
---
- hosts: localhost

  roles:
    - role: openshift-label-by-expression
      vars:
        target_namespace: labs-ci-cd      ## Apply this to ONLY objects in the specified namespace
        name_pattern: ".*selenium.*"      ## Regex patterns are supported
        label: 'app=selenium'             ## The label to be applied
```

#339 

N/A

cc: @oybed @logandonley @tylerauerbeck

### DIFF
--- a/roles/openshift-label-by-expression/README.md
+++ b/roles/openshift-label-by-expression/README.md
@@ -1,5 +1,9 @@
 # Apply a given label to OpenShift objects based on a name pattern
 
+## Overview
+This role allows you to apply a label to all objects (except Pods) which match a given expression. This is useful for applying
+labels to items which all contain a certain string. This role will OVERWRITE ALL EXISTING LABELS for those objects.
+
 ## Example:
 ```yaml
 ---
--- a/roles/openshift-label-by-expression/README.md
+++ b/roles/openshift-label-by-expression/README.md
@@ -1,0 +1,14 @@
+# Apply a given label to OpenShift objects based on a name pattern
+
+## Example:
+```yaml
+---
+- hosts: localhost
+
+  roles:
+    - role: openshift-label-by-expression
+      vars:
+        target_namespace: labs-ci-cd      ## Apply this to ONLY objects in the specified namespace
+        name_pattern: ".*selenium.*"      ## Regex patterns are supported
+        label: 'app=selenium'             ## The label to be applied
+```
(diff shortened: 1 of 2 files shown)


tylerauerbeck (dismissed):
@InfoSec812 Is this the right place for this role? Would this be better suited for something like the `openshift-applier`? I don't want to say one way or the other, but would like to see some other folks thoughts. In my opinion, this seems like something you would want to do because of something app specific, which seems a better fit for the applier.

cc/ @oybed @logandonley for some additional thoughts

oybed:
@tylerauerbeck yeah, this is the correct place (at least for now). We do not want to have roles co-existing with the openshift-applier, for a few reasons:
- The roles should *not* be `openshift-applier` specific (although their main use may be, they should be able to be re-used elsewhere)
- In the case of where the `openshift-applier` uses the roles (and it uses galaxy to pull in the repo), it has been the source for confusion before whereas the openshift-applier uses galaxy on its own repo. 

As we are looking to re-structure some of the `redhat-cop` repos, this can also be part of the discussion, but for now it make sense that openshift specific roles are part of the **container automation** (a.k.a.: CASL) repo.

makentenza:
@oybed @InfoSec812 even I don't really see the reason for having this role here (I'm the same opinion as @tylerauerbeck that this should be in another place, not openshift-applier either) does not make more sense to extend the existing `openshift-labels` role to include this new capability?

sabre1041:
@tylerauerbeck @InfoSec812 @oybed A couple thoughts:

#. Agree that it should not be targeted specifically for the `openshift-applier`, but further discussions will need to occur for all of the roles in `casl-ansible`, `infra-ansible` etc on how we plan to release these through galaxy moving forward.

I have serious issues with the logic employed by this role. Specifically:

* `oc get all` does not return all objects in a namespace, but only "the most commonly used".
* The logic for excluding pods is not correct. The `name` output uses the format `pod/*`

garethahealy:
@oybed ; was anything decided on this PR?

PR #341, opened 2018-12-01
added persistent disk for docker storage

raffaelespazzoli:
also re-established using the added disk for docker storage

#### What does this PR do?
I wanted to have the option to use persistent disk for docker storage as opposed to only ephemeral disk.
My hope was that in doing so it would be possible again to shut down the instances. 
Unfortunately this is not enough because we are using instance managers, which restart the instances.
Anyway, this is still worth having.
I also re-added the property to use the added disk for docker storage, based on the new prep steps.

cc: @redhat-cop/casl


garethahealy:
looks like this needs a review, @oybed ?

Issue #301, opened 2018-08-29
OCP on OSP v3.10 doesn't leverage static registry PV


oybed:
With OCP v3.10, the pre-allocated PV isn't utilized (it rather attaches to a dynamic PV if a storageclass is provided). This needs to be fixed.

oybed:
@bvkin has filed a BZ and found this to be an `openshift-ansible` bug. More details here:
https://bugzilla.redhat.com/show_bug.cgi?id=1657003

oybed:
This has now been fixed - closing

PR #350, opened 2019-01-08
added openshift_hosted_registry_storage_annotations to cinder storage…

bvkin:
#### What does this PR do?
Adds openshift_hosted_registry_storage_annotations variable to cinder-registry.yml. 

This is to fix the static registry bug. It works with an open [pr](https://github.com/openshift/openshift-ansible/pull/10948) in openshift-ansible. Once this is merged in the bug should be fixed, but until then you must ensure that openshift-ansible in galaxy is on the pr branch. It can be tested against an openstack deployment with `openshift_hosted_registry_storage_kind: openstack`.

#301 

cc: @redhat-cop/casl

### DIFF
--- a/playbooks/openshift/openstack/cinder-registry.yml
+++ b/playbooks/openshift/openstack/cinder-registry.yml
@@ -20,7 +20,7 @@
       openshift_hosted_registry_storage_access_modes: "{{ openshift_hosted_registry_storage_access_modes | default(\"['ReadWriteOnce']\") }}"
       openshift_hosted_registry_storage_openstack_filesystem: "{{ openshift_hosted_registry_storage_openstack_filesystem | default('ext4')}}"
       openshift_hosted_registry_storage_annotations:
-      - 'volume.beta.kubernetes.io/storage-class: standard'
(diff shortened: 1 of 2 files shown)


playbooks/openshift/openstack/cinder-registry.yml
```diff
@@ -19,6 +19,8 @@
       openshift_hosted_registry_storage_volume_size: "{{ disk.volume.size }}Gi"
       openshift_hosted_registry_storage_access_modes: "{{ openshift_hosted_registry_storage_access_modes | default(\"['ReadWriteOnce']\") }}"
       openshift_hosted_registry_storage_openstack_filesystem: "{{ openshift_hosted_registry_storage_openstack_filesystem | default('ext4')}}"
+      openshift_hosted_registry_storage_annotations:
+      - 'volume.beta.kubernetes.io/storage-class: standard'
```
oybed: May want to have the `standard` set as a optional configuration - i.e.: something like:
```
- "'volume.beta.kubernetes.io/storage-class: {{ openshift_hosted_registry_storage_class | default('standard') }}'"
```

(note: I have not tested the above, so it may have syntax errors)

bvkin: Updated. Tested this and it works with removing the inner quotes,

PR #352, opened 2019-01-18
Libvirt VM provisioning (step 1/4)

ericzolf:
#### What does this PR do?

It is the first step to get Libvirt as a working infrastructure for CASL. For the time being, it does only create the necessary VMs and instructions to setup DNS for wildcard resolution, but this is the basis for all the rest.

It is still work in progress but it would make it easier for others to help/test etc if it would be merged now, and it doesn't change anything (yet) outside of the Libvirt realm, so it shouldn't break anything. And it would make further merging also easier for me...

Follow the instructions from `docs/PROVISIONING_LIBVIRT.adoc` (this is part of the test if they are understandable, if the VMs are created and DNS works as explained in the document, you're good to go.

Only discussion on GChat.

this requires the PR https://github.com/redhat-cop/infra-ansible/pull/311 to be downloaded through galaxy for this PR to work.

cc: @redhat-cop/casl


inventory/sample.libvirt.example.com.d/inventory/group_vars/all.yml
```diff
@@ -0,0 +1,37 @@
+---
```
tylerauerbeck: I see that we're using a number of individual hostvars instead of having what the cluster "looks" like inside of `all.yml`. This looks like it works -- but seems very different from the other infra setups (aws, gcp, etc.). What I'd prefer to see is something that looks similar to:

```
libvirt_infrastructure:
   libvirt_iso: /path/to/iso
   libvirt_http_host: address of http host
   libvirt_os_variant: rhel7
   masters:
     count: 1
     cpu: 1
     memory: 2048
     name_prefix: <REPLACE WITH DESIRED NAME PREFIX FOR THE INSTANCE>
     root_volume_size: 15
     ...
```

We could also look at using a "cluster template" similar to how AWS relies on a cloud formation template and GCP relies on a deployment template. I believe there's an equivalent Libvirt component to this, so perhaps a combination of the above and this could be an option? 

Interested in seeing others thoughts here as well. I'd like to make sure we have a solid foundation here as I'm sure making changes here will require changes to the other steps in these PR's.

cc/ @oybed

tylerauerbeck:
@ericzolf Have you had a chance to look at the above comments?

ericzolf:
Sorry for the delay, I'll look in all comments (which I really appreciate) but I'm currently a bit under water. Be assured, you're not forgotten.

PR #358, opened 2019-02-01
Docker volume storage refactor

bvkin:
#### What does this PR do?
Allows for the provisioning of a cluster without using extra volumes for storing containers.

Provision a cluster and ensure that `openshift_openstack_ephemeral_volumes` is set to true within your inventory.

cc: @redhat-cop/casl


roles/container-storage-setup/tasks/main.yml
```diff
@@ -0,0 +1,30 @@
+---
+- block:
+    - name: create the docker-storage config file
+      template:
+        src: docker-storage-setup-overlayfs.j2
+        dest: /etc/sysconfig/docker-storage-setup
+        owner: root
+        group: root
+        mode: 0644
+  when:
```
oybed: This `when` check should be moved to the very bottom of the file to only do the docker check/restart if "ephemeral volume" isn't used

roles/container-storage-setup/tasks/main.yml
```diff
@@ -0,0 +1,30 @@
...
+    - openshift_openstack_ephemeral_volumes == false
```
oybed: As this role will also be called for non-openstack IaaS, we should make this variable a bit more generic - maybe something like `openshift_configure_container_storage` (then in one of the `openstack` specific playbooks - perhaps `playbooks/openshift/openstack/post-provision.yml` - set the fact correctly based on the `openshift_openstack_ephemeral_volumes` value).

roles/container-storage-setup/templates/docker-storage-setup-overlayfs.j2
```diff
@@ -0,0 +1,11 @@
+{% if docker_storage_mountpoints is defined %}
+DEVS="{{ docker_storage_mountpoints }}"
+{% else %}
+DEVS="{{ openshift_openstack_container_storage_setup.docker_dev }}"
```
oybed: same as above for these - make sure they don't have `openstack` as part of the variable names

roles/container-storage-setup/defaults/main.yml
```diff
@@ -0,0 +1,15 @@
+---
+
+openshift_docker_service_name: "docker"
+
+# container-storage-setup
+openshift_openstack_container_storage_setup:
```
oybed: See other comments re: not using `openstack` in the fact/variable name.

inventory/s1-ha.osp.example.com.d/inventory/group_vars/all.yml
```diff
@@ -26,6 +26,9 @@ dns_domain: "casl.example.com"
 dns_nameservers:
 - 192.168.1.1
 
+# Set these two to true if you wish to use vm storage
```
oybed: I believe the comment is wrong as there's only one to be set

PR #364, opened 2019-03-14
Issue 19   fix openshift label role

InfoSec812:
#### What does this PR do?
Update the openshift-label role to work properly when the namespace param is set

Given a deployed `labs-ci-cd`
And a logged in `oc` client
And the current namespace is set to `labs-ci-cd`

```
cd roles/openshift-labels/test
ansible-playbook -i inventory test.yml
```

https://github.com/rht-labs/labs-ci-cd/issues/19

cc: @redhat-cop/casl @oybed @sherl0cks @pcarney8


roles/openshift-labels/tasks/main.yml
```diff
@@ -2,11 +2,20 @@
 
 - name: "Set defaults"
   set_fact:
-    target_namespace: "{{ target_namespace | default('') }}"
+    processed_namespace: "{{ target_namespace | default('') }}"
+
+- name: Empty namespace param when not needed
+  set_fact:
+    namespace_param: ""
```
oybed: Please bring the next `set_fact` into this one,e.g.:
```
set_fact:
  namespace_param: "{{ (processed_namespace|length > 0) | ternary('-n ' + processed_namespace, '') }}"
```

roles/openshift-labels/test/test.yml
```diff
@@ -0,0 +1,18 @@
+---
```
oybed: @InfoSec812 while these tests are good for what you are working on, or when paired up with the comment/documentation in this PR, they don't necessarily make sense to keep in the role - there's just too many dependencies needed in place for these to work. I'd suggest either adding to the playbook to create the necessary projects and objects (and, if so, please make it non labs-ci-cd specific), or just don't add these tests but rather a README at the top level of this role

PR #381, opened 2019-11-12
Add note to openshift-labels role

### LINKED ISSUE #380: Update openshift-labels role documentation to indicate that it can be used to apply multiple labels
It isn't immediately clear (without digging into the role itself) that this role can be used to apply multiple labels in a single run against the objects that you are interested in. We should make this more clear in the documentation.

tylerauerbeck:
#### What does this PR do?
This updates the `openshift-labels` role README to note how to apply multiple labels to an object at once.

Look at README

resolves #380 

N/A

cc: @redhat-cop/casl

### DIFF
--- a/roles/openshift-labels/README.md
+++ b/roles/openshift-labels/README.md
@@ -14,3 +14,4 @@ tasks:
 ```
 When the `target_namespace` is ommitted, the current oc client namespace is used.
 
+This role can apply multiple labels to an object at the same time. This can be done by providing a space seperated list of labels.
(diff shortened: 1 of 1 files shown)


roles/openshift-labels/README.md
```diff
@@ -14,3 +14,4 @@ tasks:
 ```
 When the `target_namespace` is ommitted, the current oc client namespace is used.
 
+This role can apply multiple labels to an object at the same time. This can be done by providing a space seperated list of labels.
```
oybed: @tylerauerbeck good call-out. Could it be good to maybe add some examples in this README?

tylerauerbeck: Added a single and multi-label example. Let me know if you'd like something more specific than what I've got in there.

oybed: Looks great

PR #383, opened 2019-11-16
Added defaults, checks for required params, target_namespace if provided and headers for route status check

garethahealy:
#### What does this PR do?
I've updated the role to follow the simliar structure as 'openshift-replicas-ready'. I've also added a headers option, which is needed when checking auth'ed routes.

Run role as a post_step with applier.

Needed by: https://github.com/redhat-cop/pipeline-library/pull/94

cc: @redhat-cop/casl


roles/openshift-route-status/tasks/main.yml
```diff
@@ -1,31 +1,26 @@
...
-  set_fact:
-    retries: "{{ retries | default(3) }}"
+- name: "Lookup route url for {{ route }}"
+  shell: "oc get route {{ route }} {{ target_namespace }} -o jsonpath='{ .spec.host }'"
+  register: url
 
 - name: "Wait for {{ protocol }}://{{ url.stdout }} to respond with status: {{ status }}"
   uri:
     url: "{{ protocol }}://{{ url.stdout }}"
+    headers: "{{ headers | default({}) }}"
```
oybed: As you are moving much of the other parameters to the defaults/main.yml file, can you do so with `headers` as well?

roles/openshift-route-status/tasks/main.yml
```diff
@@ -1,31 +1,26 @@
...
+- name: Verify Required Parameters
+  fail:
+    msg: "'route', 'protocol' and 'status' Parameters Must be Provided!"
+  when: route is not defined or route|trim == '' or protocol is not defined or protocol|trim == '' or status is not defined or status|trim == ''
 
-- name: "Set default delay time"
+- name: "Set the target namespace option if supplied"
   set_fact:
-    delay: "{{ delay | default(5) }}"
+    target_namespace: "-n {{ namespace }}"
```
oybed: Can you please model this set_fact after how it's done in some of these other roles - i.e.: only pre-pend `-n` if not already supplied? Also, even though the roles aren't meant to be openshift-applier specific, we often use `target_namespace` as the fact name to allow for a easier use in the openshift-applier. 
https://github.com/redhat-cop/casl-ansible/blob/master/roles/openshift-labels/tasks/main.yml#L9

oybed:
@garethahealy any updates/thoughts on the above comments?

garethahealy:
@oybed ; pushed requested changes. just need to test it tomorrow.

garethahealy:
ready for review/merge.

garethahealy:
@oybed ; you good to review/merge?

roles/openshift-route-status/tasks/main.yml
```diff
@@ -1,31 +1,35 @@
...
-    delay: "{{ delay | default(5) }}"
-
-- name: "Set default number of retries"
-  set_fact:
-    retries: "{{ retries | default(3) }}"
-
 - name: "Wait for {{ protocol }}://{{ url.stdout }} to respond with status: {{ status }}"
   uri:
     url: "{{ protocol }}://{{ url.stdout }}"
+    headers: "{{ headers }}"
```
oybed: I'll approve/merge as-is, but my personal preference here would be to use the `omit` feature here - i.e.:  default(omit) - to avoid sending empty headers if not supplied.

